PR #39, opened 2016-03-01
Add .gitignore file

OrkoHunter:
Running the doctests as
```
$ python -v doctests.py
```
generates multiple byte-code compiled files which should be ignored by the version control.

PS : I can add the [full version](https://github.com/github/gitignore/blob/master/Python.gitignore) of the file (which we should) if said to.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# Byte-compiled files
+__pycache__
+*.py[cod]


norvig:
Thanks! But another contributor already submitted a .gitignore pull request (sorry I was slow to respond).

Issue #12, opened 2016-02-02
DefaultDict could front-end the native 2.5 version defaultdict


GoogleCodeExporter:
```
The collections.defaultdict class:

http://docs.python.org/library/collections.html#collections.defaultdict

could be used for Python 2.5+ rather than the method in utils.py:228

import collections
try: 
   collections.defaultdict ## set builtin introduced in 2.5
except NameError:
  class DefaultDict(dict):
  .
  .
  .




```

Original issue reported on code.google.com by `arocki...@gmail.com` on 15 Sep 2009 at 3:00

reachtarunhere:
Please review the pull request https://github.com/aimacode/aima-python/pull/42 resolving the above.

norvig:
Fixed.

PR #42, opened 2016-03-01
Fix for #12 Native defaultdict for 2.5+

reachtarunhere:
**Problem**
The following is the fix as per the suggestion in #12 which suggests 
The collections.defaultdict class http://docs.python.org/library/collections.html#collections.defaultdict
could be used for Python 2.5+ rather than the method in utils.py

**Solution**:
The DefaultDict function implementation does the above and also maintains backward compatibility taking into account the fact that the native implementation in 2.5+ requires a callable for setting default values and the original implementation made use of non callables like 0 or []. The implementation allows using both a callable or  the previous method.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -232,19 +232,32 @@ def Dict(**entries):
     """
     return entries
 
-class DefaultDict(dict):
-    """Dictionary with a default value for unknown keys."""
-    def __init__(self, default):
-        self.default = default
+try:
+    from collections import defaultdict  # Introduced in 2.5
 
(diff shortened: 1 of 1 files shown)


reachtarunhere:
Please review Pull Request for the above. https://github.com/aimacode/aima-python/pull/42

reachtarunhere:
I am closing this since I don't feel that this is relevant anymore given we are ok with breaking compatibility with for the main branch. Though it might be a good idea to add this to the legacy branch.

Issue #27, opened 2016-02-02
abstract instead of NotImplementedError


GoogleCodeExporter:
```
Many methods use "abstract" instead of raising NotImplementError when they are 
destined to be implemented by a child class.

Fun fact, upon research.  I end up in 
http://norvig.com/python-iaq.html
finding an explanation for this.

Still, I think that, to avoid confusion, it should be fixed to either 
abstract() as in the FAQ or just RaiseNotImplementedError.

Otherwise, it will raise NameError which is confusing (and abstract could 
potentially be defined globally)
```

Original issue reported on code.google.com by `alexhans...@gmail.com` on 11 Oct 2011 at 6:13

norvig:
I think "raise NotImplementedError" is the simplest, most Pythonic way to go. Another reasonable approach is the abc (abstract base class) module, but I think that's overkill here.

PR #44, opened 2016-03-01
refix for issue#27 in logic.py and utils.py

AMAN3003:
This is a fix for #27 
Created abstract method for logic.py and made a raise error methods in utils.py
at the time editing utis.py~ came please ignore it


mafiya69:
Remove the backup file `utils.py~`.

norvig:
I think, for simplicity, I'm going to go with "raise NotImplementedError"

Issue #48, opened 2016-03-02
Modifying utils.py to work with Python 3+


gokul-uf:
Code in utils.py have been implemented to fit the capabilities of python 2, as this repo is for code in python 3, we should modify this to reflect the same. I am working on this and will be sending a pull request for the same.

gokul-uf:
Question: Should we keep backward compatibility as-is or should we move it to a separate branch/repo to keep the current repo clean?

reachtarunhere:
Another important factor to consider is whether we need to do away/rename the custom implementations  for which there are native implementations in the newer version. I am specially curious about the cases where these custom implementations have different names then that of the native implementations in Python 3.

Example Case: DefaultDict (in utils.py), collections.defaultdict (in Python 3)t

norvig:
reachtarunhere is right -- we should use the now-standard implementations and remove the legacy implementations and names. So, don't use DeafultDict, use collections.defaultdict throughout.

ishaan1996:
utils.product and utils.argmin_list not working
------------------------------
utils.product([2,3,4])
NameError: name 'reduce' is not defined
------------------------------
a = ['one', 'two', 'three', 'six']
utils.argmin_list(a, len)
[]

gokul-uf:
That is the final goal. There is another branch for transistioning python 2
code to python 3. Please check that out

S. Gokula Krishnan, Fourth-year Undergraduate student
Department of Computer Science
Birla Institute of Technology & Science, Pilani
Pilani Campus, Rajasthan, INDIA - 333031

PR #53, opened 2016-03-03
Port utils.py to python3.5

tmrts:
This PR ports utils.py to python3.5 and adds py.test tests (utils_test.py).

Changes are listed below:

- In the Roadmap it was mentioned that python2 will be kept in a different branch, so I've cleaned up some compliancy code (e.g. 'try: frozenset').
- PEP8 compliant spacing
- Use of generators when necessary
- Modifying '%' operator to 'format' method.
- Using 'print' as a function
- Moved most of the doctests at the end of the module to `utils_test.py` as py.test assertions.
- Replace loops with list/gen comprehensions whenever readable.

I replaced some of the code with the stdlib code. For example, I've removed `DefaultDict` and
replaced it with `from collections import defaultdict as DefaultDict` so that the changes are readable and not breaking. I'll group imports to the beginning of the module after a code review.

Also, in some cases in order to not break the API I've added `TODO` items.

If the changes seem reasonable, I'll extend the test suite.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -1,253 +1,20 @@
 """Provide some widely useful utilities. Safe for "from utils import *".
-
 """
 
-from __future__ import generators
 import operator, math, random, copy, sys, os.path, bisect, re
-
-assert (2,5) <= sys.version_info < (3,), """\
-This code is meant for Python 2.5 through 2.7.
-You might find that the parts you care about still work in older
-Pythons or happen to work in newer ones, but you're on your own --
-edit utils.py if you want to try it."""
-
-#______________________________________________________________________________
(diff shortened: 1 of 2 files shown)


tmrts:
Also addresses #48

norvig:
Thanks tmrts -- I used your code, but I got confused about how to do a "merge" in git, and ended up using your code by copy and paste rather than the official git mechanism.

tmrts:
No problem, I'm glad it's useful :)

For future reference, don't bother with code that has conflicts just leave a note and the author should rebase it. Otherwise as the pull requests increase you'll have a hard time keeping up with the changes and not making any mistakes during the merge.

PR #63, opened 2016-03-05
Update utils.py

utk1610:
To support math function - import math module in python 3+


norvig:
Thanks for the work, but others made similar changes, and someone had the correct idea of using statistics module for mean, stddev, etc., so we don't need this pull request.

PR #66, opened 2016-03-05
TODO (completed): Move >>> doctests from utils.py to utils_test.py

SnShine:
Here are the changes:
* Moved all the >>> doctests from utils.py to utils_test.py
* Fixed bugs in argmin_list & argmax_list methods
* Modified how argmax_list & arg_max are defined. Old methods fail if the fn is len.
* Modified print statements to Python3 style in agents.py

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -54,7 +54,7 @@ class Thing(object):
 
     def show_state(self):
         "Display the agent's internal state.  Subclasses should override."
-        print "I don't know how to show_state."
+        print("I don't know how to show_state.")
 
     def display(self, canvas, x, y, width, height):
         # Do we need this?
@@ -94,7 +94,7 @@ def TraceAgent(agent):
     old_program = agent.program
     def new_program(percept):
         action = old_program(percept)
-        print '%s perceives %s and does %s' % (agent, percept, action)
+        print('%s perceives %s and does %s' % (agent, percept, action))
         return action
     agent.program = new_program
     return agent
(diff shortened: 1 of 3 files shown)


norvig:
Nice work @SnShine. I'd like to merge this, but as a github newbie, I don't see how to do it easily. Is there a way to do the merge in Github Desktop? Or do I have to issue shell commands? Or can you easily put in a new pull request against the current state of the repository?  Thanks -- I'll get the hang of this yet ...

SnShine:
Thank you.
Yeah, I will rebase my branch 'addPytests' to resolve conflicts in utils.py. Then you can merge it as a normal pull request.

SnShine:
Yeah, @norvig, I have resolved the conflicts. You can merge now as you are merging normal pull request.

norvig:
Thanks @SnShine, good  work.

tmrts:
@norvig I don't we should remove doctests of functions, because they provide clean examples. I believe it would be better if we have comprehensive tests in `*_test.py` and readable usage examples as doctests. 

Of course doctests at the end of files should be translated to proper tests.

norvig:
@tmrts and @SnShine I agree with your points

Good point. I think there are two purposes of tests: examples, and true tests. 

For functions where it is not immediately obvious how to use them, it would be good to have exactly one example in the docstring.  For example:

def product(numbers):
    """Return the product of the numbers.
    >>> product([2, 3, 4])
    24"""

What we don't want is a long collection of true tests in the docstring that cover all the corner cases (one number, no numbers, zeroes, etc.).

PR #67, opened 2016-03-05
Created Separated 2D grid environment file i.e. grid.py

utk1610:


tmrts:
Please merge this commit with #68, commits should represent a complete action and it cause problems in the history.

PR #70, opened 2016-03-05
Fixed pruning for first max node in aphabeta_search

Chipe1:
Directly using max_value instead of argmax properly prunes the first node and does more justice to the pseudo code


Chipe1:
Added #123 to fix that. Couldn't get the code to match the one in the book though. Any suggestions are appreciated.

PR #78, opened 2016-03-06
modify argmin_list and argmin_gen functions in utils.py

wolfsgang:
Rectified the functions argmin_list(seq,fn) and argmin_gen(seq,fn) . The value to be checked against should be the function value of the smallest element , not the element itself.
ex, 
>>>argmin_list(['one', 'to', 'three', 'or'], len)
>>[]

as len('to') != 'to'

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -75,11 +75,11 @@ def unique(seq):
     return list(set(seq))
 
 def product(numbers):
-    """Return the product of the numbers, e.g. product([2, 3, 10]) == 60"""
-    result = 1
-    for x in numbers:
-        result *= x
-    return result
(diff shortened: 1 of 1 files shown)


SnShine:
I think this is same as my pull request #66. Have a look at it. There is a bug in argmin_list functions and I have resolved it. Gonna update my pull request soon.

SnShine:
I don't think this is a good pull request. You have removed the functions Point(), point_x(), point_y() and fell back to the old definition of product() function.

norvig:
@SnShine is right -- things have changed, this is no longer a good pull request

PR #80, opened 2016-03-06
Enable continuous testing

tmrts:
This PR integrated travis so that each pull request is tested and linted with PEP8 and pyflake. 
This way we won't accept buggy code or revert buggy commits in the future. 

I've added the necessary configuration. All you need to do is:
1. Go to [travis-ci.org](https://travis-ci.org).
2. Sign-up using github.
3. Enable aima-python repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: 
+  - python
+
+python:
+  - "3.5"
+
+install:
+  - pip install flake8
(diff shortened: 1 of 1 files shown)


norvig:
Thanks, tmrts -- I noticed that Ciaran was doing this for aima-java and was thinking of doing it here.

PR #82, opened 2016-03-06
Add best practices for contributions

tmrts:
Adds a guide of do's and dont's for contributors about reporting issues, submitting patches and testing changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+How to Contribute to aima-python
+==========================
+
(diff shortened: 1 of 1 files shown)


Chipe1:
Shouldn't we be using pytest3 instead of pytest

Chipe1:
Shouldn't we be using "pip3 install pytest" instead of "pip install pytest"

tmrts:
@Chipe1 pip is included as a standard library module starting with python3.4. It's now used as `pip` or `python -m pip`. This shouldn't cause any confusions

PR #85, opened 2016-03-06
Add aima datasets

tmrts:
Migrates datasets from google code [aima-data](https://code.google.com/archive/p/aima-data/source/default/source) repository.

This is a temporary solution to properly test modules while refactoring them.

We can move to a better solution after @norvig creates a github repository for aima-code.
*Note* for @norvig, Creating a github repository for aima-code requires porting from svn to git, so let me know if you need any assitance


norvig:
How about if I create an empty aima-data repository, and then we can add to that?

tmrts:
The previous history would be gone. If that's not a problem, we can do what you're suggesting. Let me know when you create it

PR #88, opened 2016-03-06
Port and test text module for python 3.5

tmrts:
This PR ports `text` module to python 3.5 and adds tests for it.

**Note:** I've intentionally didn't touch a random sampling doctest, since I don't have access to the random seed used with that particular test.

### DIFF
--- a/text.py
+++ b/text.py
@@ -18,7 +18,7 @@ class UnigramTextModel(CountingProbDist):
 
     def samples(self, n):
         "Return a string of n words, random according to the model."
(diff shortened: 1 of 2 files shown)


norvig:
Doctests that produce nondeterministic output like that should probably go into a .ipynb file, not a _test.py.

tmrts:
@norvig we can override the python's RNG seed to get a deterministic output. I think there should be tests for them as well. Otherwise it's hard to find errors for those.

PR #90, opened 2016-03-06
Fixed multiple errors to make py.test rum successfully

Chipe1:
Converted dict objects to list at required places
Also changed name of class expr to Expr.
Fixed probability_test.py where it assign cpt improperly, calls methods with improper arguments and uses an unimplemented class
Fixed NameError in probability_test.py and changed xrange to range in probability.py


Chipe1:
#93 fixes all of these and more.
No need of this pr anymore

Issue #83, opened 2016-03-06
Make the Build Pass!


norvig:
Build is failing; right now because map in logic.py returns a generator, not a list. We need to finish moving to Python 3 to make the build pass.

tmrts:
There are a lot of  incompatibilities at the moment. This issue is essentially identical with #73

Chipe1:
How to make it pass?
The py.test run fine for me . Is there some other tests to be done? How do I run the complete tests since I can see most error only after sending the pull requests.

Sorry for the noob questions

SnShine:
Hey, all
I am working on this and here is my initial PR #94. Gonna add more modifications to that PR to make the whole repo pass build tests.

PR #91, opened 2016-03-06
Completed task with label #68

utk1610:
grid.py had already been created with label #68 commit.


utk1610:
Test failed because I've removed the vector_clip() function to grid,py but utils_test.py didn't imported that file.

SnShine:
hey, @utk1610, it is fine. Right now the who repo is transitioning to Py3 and there are a lot of things to breaking the build. So, it's better we do this #83 soon.

PR #94, opened 2016-03-06
Makes current repository to pass the unit tests

SnShine:
[This will fail build check now]
[More commits to come]

Note: Sorry for the long list of modifications. But these are the changed in this pull request to make the repo pass build tests. Glad I have not committed for each and every modification.
Here are the changes in this pull request:

Commit c209981
* [learning.py] modified print functions in to Py3
* [logic.py] as map no longer returns a list, modified that

Commit ecdc854
* [probability.py] as dict.key no longer supports indexing, modified it
* [probability_tests.py] removed some old tests for which there are no methods in probability.py
* [probability.py, search.py] changed deprecated xrange function

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -78,14 +78,14 @@ class DataSet:
         to not use in inputs. Attributes can be -n .. n, or an attrname.
         Also computes the list of possible values, if that wasn't done yet."""
         self.target = self.attrnum(target)
-        exclude = map(self.attrnum, exclude)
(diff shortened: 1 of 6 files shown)


SnShine:
Wow, I don't know about the tools 2to3 (Py2 project to Py3) and pass PEP8 script. @MicroT did a great job at #93.
Closing this PR now as this is intended to do what #93 has already done.

Issue #75, opened 2016-03-05
Jupyter/Ipython notebooks


norvig:
Eventually, we want to have a `filename.ipynb` for each `filename.py`, where the Notebook will give examples and documentation for how to use the code.

jeff3456:
Hi @norvig ,

I would like to take on the creation of iPython notebooks for each filename.py

What sorts of examples/ visualization should I implement? Am I free to choose the examples or should they be directly from the book?

norvig:
Anything that explains how to use the code. When there are examples from the book, great -- but not necessary.

reachtarunhere:
@norvig do you think that having gui like simulations will be useful. I think these can be helpful and fun to play with. For example in case of Chapter 2 we can allow implementing a grid world for the vacum robot. Also we can have a visualizations of different searches where user can make use of IPython's %debug to go step by step through the code and see how the frontier is building etc. If you like these ideas I would love to start working on them.

norvig:
Yes, I like this idea. I don't know enough about IPython to know if it can
be done well there, or has to be done in javascript. If you can do it in
IPython, I'm all for it.

reachtarunhere:
I guess we can do some of them without making use of JS. In some examples we might require embedding JS to add interactivity. I will give it a shot and see what I can hack around without using JS at all.

norvig:
See what you can do. I have nothing against embedding JS in Notebooks -- my
question was whether the notebook framework was so limiting that it would
be easier to have an all-javascript/html/css page without the notebook. But
if you can make things work in the notebook, that would be great.

gokul-uf:
@norvig IPython/Juypter notebooks do not provide facilities to use interactive animations out of the box. By interactive animations, I am referring to buttons .etc not dragging or rotating a 3D plot, for example. This functionality can be provided by JS. However, I have not seen examples where JS has been integrated into IPython notebooks. To ensure that code and visualizations are not separate and to allow people to play around with the notebook, we could instead use a series of static images (one for each iteration, perhaps?).

REFS:
1. http://jakevdp.github.io/blog/2013/05/12/embedding-matplotlib-animations/
2. https://jakevdp.github.io/blog/2013/06/01/ipython-notebook-javascript-python-communication/

reachtarunhere:
@gokul-uf My idea when I planned to do it without js is exactly the same of what you mentioned (changing static images). We could use embedding JS as a last resort.

norvig:
[...]

PR #98, opened 2016-03-06
Ipython created notebooks with python3 kernels

jeff3456:
As mentioned in issue #75 I created the base notebooks to show examples of the code
- I made a directory called iPython Notebooks which contains all of the notebooks
- I modified agent.py a bit so that it would compile inside the Jupyter kernel. 
- **My changes to agent.py should definitely be reviewed because I am not sure if I correctly translated from python 2 to python 3.**


jeff3456:
Hi @norvig , I accounted for the pathing issue by creating a small module (4 lines) which directly links the iPython notebooks to the parent directory, thereby being able to import anything from the parent directory. I tested it and it definitely works and can import all python files from the previous directory.

The following is the associated module. But if you feel more comfortable with have all the notebooks together in the root directory I'm fine with that also. Let me know!

```
"""This small utility is used specifically for 
   iPython Notebooks to create a import module
   path to the directory containing all of the
   algorithms, namely the parent directory.
"""

import os
import sys

cwd = os.getcwd()
sys.path.insert(0, cwd.rstrip('iPython Notebooks'))
```

jeff3456:
@norvig

PR #100, opened 2016-03-07
all  % string formated to .format is done in logic.py

Nishant23:


Nishant23:
space between 'string' and .format has been removed

PR #101, opened 2016-03-07
all % string formated to .format is done in logic.py

Nishant23:
There is no space between 'string' and .format()


norvig:
This is a good change, but there are many conflicts due to reorganization -- could you fix and resubmit?

PR #108, opened 2016-03-08
Add test cases for histogram in utils_test.py

youthinkk:
Three test cases for histogram are added. They are test_histogram_no_function(), test_histogram_with_function() and test_histogram_with_mode_one().


youthinkk:
@norvig, I have put three test cases into one under test_histogram().

PR #114, opened 2016-03-10
Rollback to one folder

MircoT:


norvig:
Thank you @MicroT, both for this change and for exploring the multi-folder approach.

PR #116, opened 2016-03-10
Implements backpropagation

Schrodinger1926:
This PR contains
- Implementation of  backpropagation (tested)
- Implementation neural net learner
- adds sigmoid and sign func to utils
- adds tests for sigmoid and sign
- pep8 changes for utils_test.py

Please let me know if perceptron learner is required separately. 
thanks

### DIFF
--- a/aimaPy/learning.py
+++ b/aimaPy/learning.py
@@ -417,24 +417,131 @@ def DecisionListLearner(dataset):
 #______________________________________________________________________________
 
 
-def NeuralNetLearner(dataset, sizes):
-    """Layered feed-forward network."""
+def NeuralNetLearner(dataset, hidden_layer_sizes=[3], learning_rate=0.01, epoches=100):
+    """
+    Layered feed-forward network.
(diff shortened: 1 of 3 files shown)


norvig:
I don't think `sign` is the right name; to me, `sign` has results +1 or -1, not +1 or 0. But it is certainly true that some authors do use `sign` or `sgn` for the function you define and I'm not sure what a better name is.

Instead of `assert 0.5-eps <= sigmoid(0) <= 0.5+eps`, let's use `math.isclose(0.5, sigmoid(0))` or optionally `from math import isclose`.

Schrodinger1926:
I got this from from old edition of aima, will correct that.

norvig:
In 3rd edition, we call that function _Threshold_, and sigmoid we call _Logistic_. In 4th edition, I'm not sure -- I'm excited that Geoff Hinton will be writing the Neural Nets / Deep Learning chapter, and we'll see what notation he prefers.

PR #119, opened 2016-03-11
Fix flake8 warnings

lucasmoura:
Fix flake8 warnings for both source and test files. The following warnings were ignored:
* F403 'from utils import *' used; unable to detect undefined names

And for line too long warnings, some were fixed when the code readability was not affected negatively.


norvig:
Thanks, @lucasmoura !

PR #120, opened 2016-03-11
modified according to pseudo code in third edition

SnShine:
Can we move this table to main readme.md in master branch?


norvig:
Good idea to move -- I accepted another pull request and moved that -- did I miss anything?

SnShine:
Everything is fine. Now readme.md looks nice.

PR #121, opened 2016-03-11
Created test_search.py

Chipe1:
Added test of breadth_first_tree_search


reachtarunhere:
@Chipe1  I have a small suggestion. This may not be relevant if you have thought about this in a different manner. So, in case if I am wrong with this please ignore this comment or better correct me. Can we use the GraphProblem class along with the Graph class instead of defining our own class in the test file? I feel that might be better as we would like to introduce minimum amount new code in our test files and rely mostly on imports and assert statements.

Chipe1:
@reachtarunhere Thanks! I didn't know that graph was already implemented. I didn't check in the middle of the code for classes. I was about to put graph for Fig 3.2. You saved me a lot of effort.

reachtarunhere:
Glad to be so some help :smile:

Issue #34, opened 2016-02-02
alpha beta pruning is incomplete


GoogleCodeExporter:
```
the use of argmax() for the first max node in the alpha beta pruning searches 
will not allow alpha pruning to be updated among the min successors of the 
initial max mode.  code in question:

    # Body of alphabeta_search:
    return argmax(game.actions(state),
                  lambda a: min_value(game.result(state, a),
                                      -infinity, infinity))

alpha will always be -infinity instead of a value discovered in a previous min 
child/successor.
```

Original issue reported on code.google.com by `chasmosa...@gmail.com` on 22 Nov 2012 at 10:16

darkcoderrises:
If no one is working on this, could I?

Chipe1:
#70

PR #123, opened 2016-03-11
Fixed alphabeta_search

Chipe1:
Tried to make it as close to the book as possible. Please comment if there is a better solution to find the action which gives the best outcome. Some solutions are:
1) Make MAX-VALUE and MIN-VALUE return even the action which lead to the value [ Different from the book ]
2) Recompute MIN-VALUE for each resultant state and compare with MAX-VALUE [ Beats the point of pruning in the first step if answer is far in the list of actions ]

### DIFF
--- a/games.py
+++ b/games.py
@@ -68,16 +68,13 @@ def alphabeta_full_search(state, game):
         return v
 
     # Body of alphabeta_search:
-    v = -infinity
+    best_score = -infinity
     best_action = None
     for a in game.actions(state):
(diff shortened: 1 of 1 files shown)


norvig:
This is good. Another way to do it is with @memoize decorator; I think your way is good for 3rd edition code; let's wait and see what 4th edition looks like to see if we need to alter this.

Chipe1:
@norvig Hope that end issue #34

PR #128, opened 2016-03-12
Improvements, corrections and tests for games.py

chiragvartak:


norvig:
Overall this looks good, but I eliminated Struct, replacing it with collections.namedtuple, which is the more modern way to do this.

Also, I note you had this comment:
update(self, h=h, v=v, k=k) # What is this exactly?

I agree with you that `update` is probably confusing. I'd be happy if we remove update; in this case it would be replaced with

```
self.h = h
self.v = v
self.k = k
```

The `update` function is more concise, but here I think explicit is better.

games.py
```diff
@@ -1,9 +1,13 @@
...
 """
 
+import collections
+import math
+import random
+
 from utils import *  # noqa
 
-import random
+infinity = math.inf
```
Chipe1: utils already has infinity defined

chiragvartak: It was removed in a recent PR.

chiragvartak:
Sir, you are looking at an older commit. Please take a look at the latest commit, the final PR. It passes all tests and has no conflicts with the base branch.

games.py
Chipe1: @norvig Was that a mistake? Because I think infinity is a nice variable to have in utils.py
I can't think of any reason to remove it.

PR #129, opened 2016-03-13
Revised Implementation of backpropagation, util func

Schrodinger1926:


Schrodinger1926:
@norvig Travis is giving an unusual error while installing python3.5 on server, Please trigger the build again.

PR #134, opened 2016-03-13
IPython Notebook for games.py

chiragvartak:
I have created a notebook for the module `games.py`. @norvig , it would be great if you could have a look at it.  
Comments, suggestions or feedback in any way on this notebook would be great! I have been quite informal in the language I have used. I hope its acceptable, but please tell me if it isn't.  
If viewing the notebook is a problem because this is a pull request, it can be viewed here: [Nbviewer online link](http://nbviewer.jupyter.org/github/chiragvartak/aima-python/blob/games-ipython-notebook/games.ipynb)


norvig:
This looks very nice. I think the informal tone is just right -- there is plenty of formal language in the textbook; to be valuable, the notebooks should be a different way of learning.

@chiragvartak , how about taking the first paragraphs out of `games.ipynb` and put them in a new `intro.ipynb`, explaining how notebooks work, how to install them, and what to expect throughout aima-python?

chiragvartak:
Thank you sir, glad that it's okay. :)  
Yes, I like the idea of an `intro.ipynb` notebook. I'll start on getting a first draft of this ready.

PR #142, opened 2016-03-16
modified index of code to reflex changes in 3rd edition of the book

SnShine:


SnShine:
Hey, @norvig, I just realised I am using Indian subcontinent version of the 3rd edition published by Dorling Kindersley (India) Pvt. Ltd. I don't think these changes would be useful for people using 3rd edition published by other publishers.
Can I know how many versions are there for the 3rd edition. And what do you suggest regarding this issue, page numbers of pseudo codes are different in different versions?

norvig:
Right @SnShine , page numbers can be slightly different. But section numbers should be the same. Let's drop the page numbers and just use section numbers.

PR #144, opened 2016-03-16
Implemented forward-backward function in probability.py

SnShine:
Changes:
* Added element_wise_product method to utils.py.
* While implementing forward_backward, I have used a new class HiddenMarkovModel. We can also use already implemented BayesNet class, but I think this implementation is pretty straight forward.
* Added tests for element_wise_product and forward_backward methods

And, @norvig, what have we decided regarding the function `update()`? I think using `self.a= a` is better than using `update(self, a=a)`.

### DIFF
--- a/probability.py
+++ b/probability.py
@@ -524,15 +524,7 @@ def markov_blanket_sample(X, e, bn):
 
 # _________________________________________________________________________
 
-"""
-umbrella_evidence = [T, T, F, T, T]
-umbrella_prior = [0.5, 0.5]
-umbrella_transition = [[0.7, 0.3], [0.3, 0.7]]
-umbrella_sensor = [[0.9, 0.2], [0.1, 0.8]]
-umbrellaHMM = HiddenMarkovModel(umbrella_transition, umbrella_sensor)
-
-print(forward_backward(umbrellaHMM, umbrella_evidence, umbrella_prior))
-"""
+# Umbrella Example [Fig. 15.2]
 
 class HiddenMarkovModel:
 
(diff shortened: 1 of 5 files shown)


utils.py
```diff
@@ -162,11 +162,17 @@ def histogram(values, mode=0, bin_function=None):
         return sorted(bins.items())
 
 
-def dotproduct(X, Y):
+def sum_dotproduct(X, Y):
```
Schrodinger1926: I think dotproduct should be fine

utils.py
```diff
@@ -162,11 +162,17 @@ def histogram(values, mode=0, bin_function=None):
...
     """Return the sum of the element-wise product of vectors x and y."""
     return sum(x * y for x, y in zip(X, Y))
 
 
+def dotproduct(X, Y):
```
Schrodinger1926: element wise multiplication and dotproduct are not the same thing.

Schrodinger1926:
I too was writing forward-backward :) , Anyways nice work

probability.py
```diff
@@ -524,11 +524,76 @@ def markov_blanket_sample(X, e, bn):
...
+    b = [1.0, 1.0]
+    bv = [b]    # we don't need bv; but we will have a list of all backward messages here
+    sv = [[0, 0] for i in range(len(ev))]
 
+    fv[0] = prior
+
+    for i in range(1, t+ 1):
+        fv[i] = forward(HMM, fv[i- 1], ev[i])
+    for i in range(t, -1, -1):
+        sv[i- 1] = normalize(dotproduct(fv[i], b))
```
Schrodinger1926: isn't normalize a class method?

SnShine:
Oops, my bad. What was I thinking?
@Schrodinger1926 Yeah, it's alright.

utils.py
norvig: `dotproduct` is a good name. See https://en.wikipedia.org/wiki/Dot_product

PR #145, opened 2016-03-16
Fixed a bug and added tests for logic.py

Chipe1:


norvig:
Good additions; thanks, @Chipe1 !

PR #147, opened 2016-03-16
Modify index of code in ready.md

SnShine:
Changes:
* removes page numbers (https://github.com/aimacode/aima-python/pull/142#issuecomment-197199121)
* adds all the pseudo codes in 3rd edition

Added "*" to some Figure names as I couldn't find them in the book. Let's recheck and remove/rename those rows.


norvig:
Nice one, @SnShine ; thanks!

Issue #149, opened 2016-03-17
11.5 Hierarchical-Search


MaVericKWareZ:
Missing implementation of Figure 11.5 Hierarchical-Search.
Added this issue to let you all know that i will be working on it.
Thanks !

MaVericKWareZ:
Hey @norvig Sorry for the question, but do we have anything similar t the Action subclass in java here?
Thanks!

anuj1995:
Hey @norvig i want to work on it

MaVericKWareZ:
@anuj1995 I already have the code in place for the algo.
Only the tests need to be implemented.
Maybe you could help me with that if possible?

anuj1995:
I'll like to help you in that

MaVericKWareZ:
@anuj1995  Any progress on the issue?
Sorry i've been sidetracked a bit and couldn't work on this during the last few days

hackboxlive:
Is it still open? I would like to work upon it!

dmeoli:
Section 11.5 hierarchical planning is implemented in `planning.py`: `HLA`, `RealWorldPlanningProblem.refinements`, `hierarchical_search` (Fig 11.5), `angelic_search` (Fig 11.8), `AngelicHLA`/`angelic_action`, `reach_opt`/`reach_pes`/`find_reachable_set`/`making_progress`. It's covered by `test_refinements`, `test_hierarchical_search` (+ `_max_depth`), `test_angelic_search`, `test_is_primitive`, `test_making_progress`, and the reachable-set tests, all passing. Closing as implemented.

PR #154, opened 2016-03-18
Add HierarchicalSearch

MaVericKWareZ:
Implementation of Issue #149

### DIFF
--- a/planning.py
+++ b/planning.py
@@ -1,9 +1,9 @@
 """Planning (Chapters 10-11)
 """
-
+from logic import Expr
 from utils import *
 import agents
-from search import Node
+from search import *
 
 import math
 import random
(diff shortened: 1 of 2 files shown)


planning.py
```diff
@@ -12,3 +11,67 @@
...
+    The initial plan supplied to the algorithm is [Act]. The REFINEMENTS function 
+    returns a set of action sequences, one for each refinement of the HLA whose preconditions
+    are satisfied by the specified state, outcome.
+    author :Sarthak Mahapatra"""
+
+    current=Node(problem.initial)
+    frontier=FIFOQueue()
+    frontier.append(current)
+    while True:
+        if frontier.__len__()==0:
```
norvig: use "if not frontier"

planning.py
```diff
@@ -12,3 +11,67 @@
...
+            if outcome statisfies.problem.GOAL then return plan
+        else for each sequence in REFINEMENTS(hla,outcome,hierarchy)do
+            frontier <-INSERT(APPEND(prefix,sequence,suffix),frontier)
+    Figure 11.5 A breadth-first implementation of hierarchical forward planning search.
+    The initial plan supplied to the algorithm is [Act]. The REFINEMENTS function 
+    returns a set of action sequences, one for each refinement of the HLA whose preconditions
+    are satisfied by the specified state, outcome.
+    author :Sarthak Mahapatra"""
+
+    current=Node(problem.initial)
```
norvig: Spaces around "="

planning.py
```diff
@@ -12,3 +11,67 @@
 import time
 import bisect
 import string
+from test.test_asyncio.test_events import noop
```
norvig: I'm not sure what this is for.

norvig:
You've copied some of the pseudocode, but you haven't defined what a HLA is, nor a HLA library, nor how a HLA is replaced by a refinement, nor shown an example of the code actually solving a problem. Good start, but needs a lot more work.

MaVericKWareZ:
@norvig Thanks for the feedback Sir!.
I will be working on this to rectify the mistakes you've mentioned.
Just a question, Do i have to implement a Action class for a HLA or is their some other preferred way?Need your inputs here. 
Thanks!!

planning.py
MaVericKWareZ: This one is entirely my fault. Forgot to Cleanup the code before submitting the PR.
Thus these unnecessary lines.

norvig:
Good questions, @MaVericKWareZ , and I'm glad you're sticking with this. In Figure 11.4 from https://github.com/aimacode/aima-pseudocode/blob/master/algorithms.pdf we see that a Refinement has an HLA, a list of steps, and optionally a precondition. The HLA has an operator and arguments, and can have variables. You could use the Expr class from logic.py for HLAs; probably a subclass of Expr.

MaVericKWareZ:
@norvig Sir,Please do have a look at the Code now.
Am i moving in the right direction? Awaiting your feedback and further inputs on this issue.
Also i have tried to add the test_planning.py file for the pytests required for hierarchicalSearch().

MaVericKWareZ:
There is some prob with my fork of aima-python.So i a closing this PR.
Will follow up with an updated one.

PR #155, opened 2016-03-19
modified small typo in logic.py

SnShine:


SnShine:
No need of this as I have added this in #156.

PR #156, opened 2016-03-19
Made the output of play_game() a bit more verbose

SnShine:
Changes:
* Prints state of the game after each player played their turn.
* Notebooks are supposed to be clear without any outputs from the execution of the code cells. Cleared the outputs in all notebooks. So, everyone, before committing notebooks, just restart the kernel by clearing all the outputs.
* fixed a typo in logic.py

And @norvig, I am trying to implement interactive TicTacToe game in notebooks. In our `query_player()` function, we have `input()` which takes the input move from the player. Shall we change the way how we take the input from user? Because in notebooks, we need to take input as on_click_event() on any object and state shouldn't be displayed as it is displaying now via `game.display(state)`. Or shall I implement another method if we are running the game from inside notebooks. In the later case we can play the game both in command line and in notebooks (interactively!). Need your inputs here.

### DIFF
--- a/agents.ipynb
+++ b/agents.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {
     "collapsed": false
    },
--- a/csp.ipynb
+++ b/csp.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {
     "collapsed": false
    },
--- a/games.ipynb
+++ b/games.ipynb
@@ -594,7 +594,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.5.0"
+   "version": "3.5.1"
   }
  },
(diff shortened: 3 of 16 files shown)


chiragvartak:
@SnShine, it makes little sense to remove the outputs from the notebooks. We cannot predict how the readers plan to use these IPython Notebooks. There will be quite a few readers who won't install the necessary apps to download and run the Notebooks, and will just prefer reading them online in their browsers. Also, these notebooks are very portable and can to converted to a variety of formats, including static HTML pages, Latex documents, Python code and whatnot. A reader not interactively running these notebooks will then, be able to make little use of them.  
In any case, removing the outputs would make it difficult for these readers to learn from the notebooks. Also, its always better having the outputs rather than not having them, correct? Any one who is bothered by them can simple clear them in their notebook app. :)

SnShine:
I always think it is good to have notebooks free from outputs so user can run them and learn from them while doing so.
But it would make easier to people to convert notebooks to other formats if there are outputs in the notebooks. Viewing notebooks online would surely requires outputs to be present.

norvig:
I think we should generate and store notebooks with all output. Then, a user (or we) can clear the output for use when we know an interactive notebook is available. 

I think there can be more than one "human" player -- one that produces moves by typing input, and one that clicks.

PR #157, opened 2016-03-19
implemented fixed_lag_smoothing

SnShine:
Adds:
* matrix_multiplication in utils.py and tests for it
* implements fixed_lag_smoothing and tests for it in probability.py

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -172,28 +172,28 @@ def element_wise_product(X, Y):
     assert len(X) == len(Y)
     return(list(x * y for x, y in zip(X, Y)))
 
-def _mat_mult(X_M, Y_M):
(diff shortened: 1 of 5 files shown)


utils.py
```diff
@@ -163,15 +163,43 @@ def histogram(values, mode=0, bin_function=None):
...
     return sum(x * y for x, y in zip(X, Y))
 
 
 def element_wise_product(X, Y):
-    """Return vector as an element-wise product of vectors x and y"""
+    """Return vector as an element-wise product of vectors X and Y"""
     assert len(X) == len(Y)
     return(list(x * y for x, y in zip(X, Y)))
 
+def _mat_mult(X_M, Y_M):
```
Chipe1: Would be better if you moved this into the matrix_multiplication function

PR #161, opened 2016-03-22
Fixed ussue where an agent's location set to 0 returned None

tolusalako:
I was implementing a simple agent for an ipython example, and decided to use a 1-Dimensional location starting at 0.

`thing.location = location or self.default_location(thing)`

0 or'ed with None returned None.

### DIFF
--- a/agents.ipynb
+++ b/agents.ipynb
@@ -1,14 +1,236 @@
 {
  "cells": [
   {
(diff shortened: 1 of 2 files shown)


norvig:
Very good, @tolusalako -- I was thinking of two-dimensional locations, which can never be false, but you are right to check only against `None`.

PR #163, opened 2016-03-22
Add tests for the Wumpus World

chiragvartak:
This PR adds a few tests created to the test the module logic.py w.r.t. the situation depicted in Fig. 7.2., and explained in more detail in Sec. 7.4.3.


logic.py
```diff
@@ -81,9 +82,10 @@ def tell(self, sentence):
         self.clauses.extend(conjuncts(to_cnf(sentence)))
 
     def ask_generator(self, query):
-        "Yield the empty substitution if KB implies query; else nothing."
+        "Return the empty substitution {} if KB entails query; else return False."
```
Chipe1: The return value is None, not False

logic.py
```diff
@@ -81,9 +82,10 @@ def tell(self, sentence):
...
         if tt_entails(Expr('&', *self.clauses), query):
-            yield {}
+            yield {}  # Why use yield when you are not returning a generator?
```
Chipe1: ask_generator is expected to return a generator(an iterable is also fine) for other methods to use. In this case there is only one element returned by the generator function. I think the comments must be a bit more explicit about that.

chiragvartak: For First Order Logic, yes. But for Propositional Logic, `ask_generator` will return either `{}` or `None`, which we can change to return simply `True` or `False` instead. I was wondering initially as to the original author's intention regarding this. But I think I get it now: this behavior of the function is tolerated to, maybe, maintain consistency.

norvig: That's right, @chiragvartak -- ask_generator is designed for the question "tell me all the ways that this query could be true". Each way is a substitution. When there are no variables, it either yields the empty substitution once, saying there is one way to make the query true, or it doesn't yield anything at all, meaning there are no ways to make the query true. And I wanted to have the same interface for every kind of KB. I guess we could also add

     def ask_if_true(self, query): return self.ask(query) is not False

chiragvartak: Yes, that sounds like a good idea. So, we can encourage a first-time reader who is just getting started with Propositional Logic to use `ask_if_true` instead of `ask` or `ask_generator` so that she doesn't get confused. I'll make the necessary changes to the PR.

logic.py
norvig: @Chipe1 is right -- ask_generator does not return False when it fails, it just refuses to yield any substitutions.

chiragvartak:
I have corrected the mistake and added the `ask_if_true` method. This PR should be good to go. :)

norvig:
OK. For "fix the precedence", here's the problem: it is a really bad idea to make (A == B) be a constructor that creates Expr('==', A, B), because then we get unintended effects whenever we write code that manipulates Expr's, for example, if `variables` is a non-empty list of variables, then `X in variables` would always be True, because `Expr('==', X, variables[0])` is true. So we need to pick some other operator to overload equality and inequality. Perhaps there are better ones than % and ^ -- I don't know.

For implications, `==>` and `<==`, I chose `<<` and `>>`, which also are not the right precedences. Perhaps it would be better to use `>=` and `<=`. These have the "right" precedence, but `>=` would be prettier if it were `=>` (but `expr` could handle that). And if we used these for implication, then if we ever wanted to write statements about arithmetic, we'd have to use `LT(A, B)` rather than `A <= B`. Maybe that's ok.

Issue #73, opened 2016-03-05
Port to Python 3


norvig:
We need to update all the files to Python 3. Issues include: print is now a function; range and map and other functions no longer produce lists; objects of different types can no longer be compared with <; strings are now Unicode; it would be nice to move % string formating to .format; there is a new next function for generators; integer division now returns a float; we can now use set literals; legacy functions such as `every` should be replaced with Python 3 functions such as `all`, or by generator expressions or comprehensions.

light94:
Hi @norvig ,
I want to work on this if no one has taken this up yet.

norvig:
Great work; the code now passes Python 3 build/test. But there are some outdated idioms that should be replaced eventually.

anuj1995:
I want to work on it

rusrushal13:
Is this issue is still Open?

SnShine:
Nope. This problem's solved already.

PR #168, opened 2016-03-24
Fixed Outdated Idioms in Python 3 Port

larry-he:
Additional fixes for Issue #73

There were a few outdated idioms that were found using the optional fixers in 2to3. The issues fixed are as follows:

Replaced cmp: 1 instance in line 323 of search.py
Convert .sort() to heapq: 1 instance in line 150 of text.py

I also changed most uses of "doc" to "docid" in text.py. This is because the variable name "doc" falsely implies the variable is equal to the document's text when, in reality, it is equal to the index of the document. The variable name "docid" better represents this idea.

### DIFF
--- a/grid.py
+++ b/grid.py
@@ -44,4 +44,4 @@ def vector_clip(vector, lowest, highest):
     """Return vector, except if any element is less than the corresponding
     value of lowest or more than the corresponding value of highest, clip to
     those values."""
-    return type(vector)(map(clip, vector, lowest, highest))
+    return type(vector)(list(map(clip, vector, lowest, highest)))
--- a/grid.py
+++ b/grid.py
@@ -44,4 +44,4 @@ def vector_clip(vector, lowest, highest):
     """Return vector, except if any element is less than the corresponding
     value of lowest or more than the corresponding value of highest, clip to
     those values."""
-    return type(vector)(list(map(clip, vector, lowest, highest)))
+    return type(vector)(map(clip, vector, lowest, highest))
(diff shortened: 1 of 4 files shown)


SnShine:
The build is failing. You are initialising a dict with `{}` by replacing `set()` not a set.

larry-he:
Thanks for the heads up. Everything should be fine now.

grid.py
```diff
@@ -44,4 +44,4 @@ def vector_clip(vector, lowest, highest):
     """Return vector, except if any element is less than the corresponding
     value of lowest or more than the corresponding value of highest, clip to
     those values."""
-    return type(vector)(map(clip, vector, lowest, highest))
+    return type(vector)(list(map(clip, vector, lowest, highest)))
```
norvig: This `list` is not necessary, because the result of the map is already being cast to `type(vector)`

text.py
```diff
@@ -147,8 +147,7 @@ def query(self, query_text, n=10):
         qwords = [w for w in words(query_text) if w not in self.stopwords]
         shortest = argmin(qwords, lambda w: len(self.index[w]))
         docs = self.index[shortest]
-        results = [(sum([self.score(w, d) for w in qwords]), d) for d in docs]
-        results.sort()
+        results = sorted([(sum([self.score(w, d) for w in qwords]), d) for d in docs])
```
norvig: Your change is an improvement, I agree, but I would probably do it like this:

    def total_score(doc): return sum(self.score(w, doc) for w in words)
    return heapq.nlargest((total_score(doc), doc) for doc in docs)

larry-he:
@norvig Thanks for the suggestions. I've incorporated them into the pull request.

PR #170, opened 2016-03-24
Added functions to Plot Utility vs Iteration Graphs

reachtarunhere:
Matplotlib versions of Fig 23.1a Fig 25.1a in the book.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+matplotlib == 1.3.1
--- a/rl.py
+++ b/rl.py
@@ -54,3 +54,32 @@ class PassiveTDAgent:
         ''' To be overriden in most cases. The default case
         assumes th percept to be of type (state, reward)'''
         return percept
+
+
+def run_single_trial(agent_program, mdp):
+    ''' Execute trial for given agent_program
+    and mdp. mdp should be an instance of subclass
+    of mdp.MDP '''
+
+    def take_single_action(mdp, s, a):
+        '''
+        Selects outcome of taking action a
+        in state s. Weighted Sampling.
+        '''
+        x = random.uniform(0, 1)
+        cumulative_probability = 0.0
+        for probabilty_state in mdp.T(s, a):
+            probabilty, state = probabilty_state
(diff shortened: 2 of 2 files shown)


reachtarunhere:
This is also compatible with IPy notebooks. I am also working on adding usage instructions and examples of the same in the notebook. I will make a separate PR for that.

norvig:
Interesting idea. What I'm not sure of is this: how many students will have a Python 3.4+ environment already set up for them, but will not have matplotlib? Four years ago, it was quite common to have a Phython but not have matplotlib. If that is still common today, then I would prefer the plotting functions to only be in the ipython notebook.  But if we feel almost all students will have a "batteries included" version with matplotlib already loaded, then this is good. Does anyone know the answer to this question.

reachtarunhere:
It is still not common to have matplotlib installed. But the situation today is better in terms of the fact that there are lots of "batteries included" distributions available that can be installed very easily For example https://www.continuum.io/downloads . Also it might give us more flexibility if we consider making some additional packages a requirement. For example we can make nice representations of graphs/trees using networkx package in IPython. Without networkx or some similar package the same task would be more complex to achieve. People will still need to install those packages to use the notebooks.

We can add instruction in the main readme for running pip install -r requirements.txt or point people to batteries included distributions.

@norvig on a separate note regarding proposals as we are currently not aware about the changes in 4th edition do we still need to include a timeline like most conventional gsoc proposals?

norvig:
No, I decided I don't want to have the pylab dependency in the rl.py file -- let's keep that only in the rl.ipynb file. It is true that if you have ipython installed, then this will probably work. But it can also lead to confusion -- in my Anaconda Python 3.5, I can import matplotlib, but if I then try to plit, nothing happens. It is just easier to have all plotting in the .ipynb only.

Your proposal was fine; thanks for submitting, and for all your work so far.

reachtarunhere:
I will make a separate PR where the map functionality is only a part of the IPy notebooks. I will keep the iteration functionality in the rl.py files as this can also be used for text output.

PR #172, opened 2016-03-24
Genetic algorithm optimised

anuj1995:
len() function was called iteratively put it outside to avoid unnecessary calling of len()


csp.py
```diff
@@ -517,8 +517,8 @@ def display(self, assignment):
 
 def flatten(seqs): return sum(seqs, [])
 
-easy1 = '..3.2.6..9..3.5..1..18.64....81.29..7.......8..67.82....26.95..8..2.3..9..5.1.3..'  # noqa
```
norvig: I don't understand what these #noqa lines are for -- these lines are under 100 chars.

probability.py
```diff
@@ -549,7 +549,8 @@ def forward(HMM, fv, ev):
                             scalar_vector_product(fv[1], HMM.transition_model[1]))
     sensor_dist = HMM.sensor_dist(ev)
 
-    return([float("{0:.4f}".format(i)) for i in normalize(element_wise_product(sensor_dist, prediction))])
```
norvig: Not your fault, @anuj1995 , but I think this line should be changed to just return the normalized results; then somewhere else we can have a function to truncate to 4 decimal places, if we find that more convenient to look at.

chiragvartak:
`__pycache__/grid.cpython-34.pyc` and  
`__pycache__/utils.cpython-34.pyc`  
Why are these .pyc files getting committed here? We do have a `*.pyc` and `__pycache__/` in our `.gitignore` file, and it hasn't been modified in this PR. Am I missing something?

anuj1995:
#noqa is avoid unnecessary warnings

csp.py
anuj1995: #noqa is added to avoid the unnecessary warnings in flake

search.py
```diff
@@ -477,7 +477,8 @@ def genetic_algorithm(population, fitness_fn, ngen=1000, pmut=0.1):
     "[Fig. 4.8]"
     for i in range(ngen):
         new_population = []
-        for i in len(population):
+        len_population = len(population)
+        for i in len_population:
```
Chipe1: Shouldn't it be for `i in range(len(population))`
Also the length is not calculated in every iteration dynamically. The length is calculated once and used throughout the loop.

anuj1995: no, the len() will be called in each iteration returning the same value again and again

Chipe1: @anuj1995 Do you mean the ngen loop or the population loop? I was talking about the population loop. If we change it to range(len_population), evaluation of len(population) should be done before the `for i in range(ngen)` loop(assuming that the population size won't change through the generations)

anuj1995: I am talking about population loop

norvig:
I don't think this pull request is moving us in the right direction, so I'm going to close it. Computing len(population) takes a trivial amount of time; no need for a change.

PR #173, opened 2016-03-25
IPython Notebook for logic.py

chiragvartak:
I have added some content explaining a few particulars of the `logic.py` module. Of course, the notebook is not yet complete. It will be a long way till the `logic.py` notebook is complete. I have started with Propositional Logic. Eventually the notebook will have explanations of algorithms from chapters 7, 8 and 9.  
Anyone, feel free to continue writing this notebook if you wish to. I am going to take a break from Logic and work on something else for a change.  
I had a question: What is the propositional symbol for Inequality? And when do we say that it's an Inequality? For example, the propositional symbol for Implication is the right arrow (&rarr;). And it is used when the sentence in plain English is like "If A then B". What is corresponding symbol and plain English equivalent for Inequality?  
The [Nbviewer link for the notebook][1] in case you cannot view it because this is a PR.

[1]: http://nbviewer.jupyter.org/github/chiragvartak/aima-python/blob/logic-ipython-notebook/logic.ipynb


norvig:
Not bad -- thanks for doing this. Still a way to go, but this is a good start. The Nbviewer link is very helpful, thanks! I think the Introduction section can be shorter, and mostly point to the intro.ipynb.

Issue #160, opened 2016-03-21
Need some help implementing interactive TicTacToe game


SnShine:
Hey everyone,

I am trying to implement interactive TTT game and stuck at a problem. I have completed till this far (Taking user input by clicking on blocks).
![](http://g.recordit.co/UD9fWNUnZ9.gif)

You can view the source at https://github.com/SnShine/aima-python/commits/addInteractiveTTT (last three commits).
Is there any way to execute a python method after getting user_input from html onclick event. I can execute python commands from Python.notebook.kernel but unable to pass methods here.

Any idea how to do it? TIA

reachtarunhere:
Yes @SnShine my code exactly solves that problem. There are two issues that need to be handled for this. I can surely help you in this once @norvig  can tell me if what I have done is ok or not.

SnShine:
Yeah, it's alright. I haven't played with your implementation of interactive blocks yesterday. Now that I have installed ipythonblocks, I figured it out.
And once interactive blocks is finalised by Norvig, we can build TTTgrid, searching algos and stuff like that on your implementation.

reachtarunhere:
@SnShine  yeah that would be great. I will try to add a couple of more demos today. They should serve as good examples if we are going to use interactive blocks in the future.

Chipe1:
@reachtarunhere How do you plan to implement structures that are not grids. Can we use ipythonblocks to show graphs like the map of Romania given in the book and do dfs on it.

SnShine:
@Chipe1 I think we can't use ipythonblocks to represent graphs like Romania. If we are visualising that graph, we need have places separated to each other to scale. We can't do that with ipythonblocks. We can always use graphvix or some other library if we want to implement graphs like that. But, do we need to interactive with Romania map? We can have visuals for different searching algorithms, but there is no need of interact with the map/graph itself.

norvig:
I have a question about ipythonblocks: I know I can have a grid of blocks, and specify colors. But I think we'll need to be able to put text labels on blocks for many applications, and for other applications, we'd want each block to be a canvas, or at least have the option of having a block display an image. I don't think ipythonblocks can do that? If not, it might be better to find a package that can.

norvig:
@SnShine makes a good comment: I think there are three types of visualizations:

- Static: still pictures, generated by a program, like a picture of a graph.
- Animations: a series of static picture snapshots, with buttons for play/stop/rewind, but no interaction with the content, like the sequence of a state space search.
[...]

PR #174, opened 2016-03-25
Added TicTacToe to canvas

Chipe1:
Made a Canvas class based on @reachtarunhere 's continuous world  https://github.com/reachtarunhere/aima-demo. This class is a more general version which can be used to make a canvas and draw on it and interact with it using the object's methods. Currently you can draw lines, rectangle and arcs with different colors. Canvas can interact with mouse clicks. If @norvig is ok with this kind of implementation, I will add more support like text, images, mouse movement etc.


reachtarunhere:
@Chipe1  Most humbly and respectfully I want to convey that you seem to have extended part of the code which itself was waiting approval as per my previous comment. As per discussion on the thread, it would be nice if we can complete a generic class before implementing specific examples e.g. Tic Tac Toe. Hence I was working in parallel to complete it. I have implemented support for images, smoothing of walls, discrete version of the world.

I had mentioned 

>  Also I am a little confused about how this should be organized like a MVC and where this code should go when I make a PR.

which clearly conveys that I was working on putting on a PR. Frankly I don't know how we should deal with this moving forward. We should work collectively to avoid duplication of effort.I acknowledge the fact that a TicTacToe is different from a generic class. I respect your gesture for mentioning the source link but it would be better if we can avoid such situations in future.

SnShine:
@reachtarunhere Yeah, I agree with you. I have stopped developing TTT once I have had a look at your aima-demo & interactive blocks code. I am planning on to build stuff upon it.
So, best thing is open a PR with the code you have completed till now and we all build/develop upon it.

Chipe1:
My intention was to work on a class which was not related to the algorithms in the book, but to directly draw on the canvas using basic shapes and I have mentioned this in issue #160. This implementation doesn't intend to solve the MVC problem, but instead to help in the 'View' part. This will just replace the canvas part of the continuous world not the main python code. I didn't know you were working on the canvas part(eg. adding images). I thought you were working with grids to implement the walls and how to organize the MVC structure. In any case I'm sure we would need to draw basic shapes and take input from the canvas, hence this PR. I mentioned your continuous world because I couldn't figure out how to make display(HTML(...)) affect the canvas when called from kernel.execute. Your code is where I found the solution and made its structure similar to @reachtarunhere 's continuous so that integrating it into the MVC would become easy. The TicTacToe is just a rough demo made by hard-coding the methods of the class. I included that because I thought @SnShine could complete the TicTacToe in the jupyter notebook using the canvas and I could have a review if it had enough functionality. I actually wanted to add this for showing the different steps in search algorithms in search.ipynb.

reachtarunhere:
@SnShine the difficulty is that there is no clarity in terms of how the view and the model will be kept separated and where these general classes for canvas and grids go (new files? grid.py?). I asked the same question on the issue that was raised by you. Once there is a reply by the maintainers about that I'll make a PR.

norvig:
@reachtarunhere and others, my apologies for being out of the loop for a few days,

I think grid.py is a good place to put the code for managing both discrete and continuous two-dimensional worlds. There will also have to be a grid.js or canvas.js file; whatever name makes sense.

I think the model is a class that contains a mapping from locations to objects in that locations, for example {(1, 2): [Wumpus], (2, 3): [Gold, Agent]}. The View then has everything we need to draw these objects.

What else would you like to have decided? I'm excited to see this all working soon!

reachtarunhere:
@norvig Here a few things on which I need your opinion. Before the questions I would like to explain the current work-flow when both the view and model reside in one class. This is just an example.

We init the world using. a = GridWorld() and this class has methods like handle_click, handle_drag etc. So when we click a command like a.handle_click(coordinates) is called in the background and the changes are made to the model. We also have methods like a.change_background(coordinates, color) which are accessed both directly by the user in IPy shell and by our handler methods. 

1. Where do the methods that handle interactions go (view or model)?
2. How do the model and view talk to each other? This is necessary because interactions result in changes in the model. (both will need to have references/pointers to each other?)
3. Also, what happens when something like a = GridWorld() is executed. Are we initializing the model or the view. If one of them initializes the other then does the view initialize the model or other way round? I am not sure about what object (view or model) will be stored in "a". As we currently use some methods that belong to view like  a.show() and some that belong to the model like a.add_agent(coordinates).

In case if the questions are not clear please comment here and I will try to elaborate on them.

Chipe1:
@norvig, about this PR, do we need a class to manage drawing onto the canvas. With this class, we can call python functions like `object.line()` or `object.image()` to draw onto the canvas. It is useful when we do all our computation and data storage in python. This class doesn't facilitate storing any data in JavaScript as such an implementation would probably need JavaScript dependent functions and may clutter the code base. If you decide to use this class where should this class be present? It is currently in a separate .py file due to IPython import statement which works on jupyter notebook, but not in default python. Should there be a separate .js file which is easy to edit and extend or should all the JS functions be stored as strings inside the python Canvas class?

norvig:
@Chipe1 I think it would be better with a separate .js file. I like the idea of a python class that can draw lines, images, text.

reachtarunhere:
@norvig  Please answer my query. I am waiting for your reply to move forward with the PR for grid and continuous  world implementations.

norvig:
@reachtarunhere, here's an attempt to answer your questions. There are many ways to do it, but this should give you an idea. Let's take the tic-tac-toe-game as an example.

- Create the model: `tttt = TicTacToe()`. This is an object that contains the board (initially empty), the player whose turn it is (initially, say, X), and maybe an indication of whether anyone has won the game. There should be a method for `ttt.make_move(location)`.

- Create the view: `view = TicTacToeView(ttt)`. There should be  `onclick` methods that call `ttt.make_move(location)` for each location, and maybe something that changes a display when someone wins or there is a draw.

- There are two ways for the view to display itself. One is that whenever there is an `onclick` action, it queries the complete state of the view, and displays that. For small problems like tic-tac-toe, that will work fine. For more complicated worlds, we would want to do an incremental display, where the model would have to tell the view whenever something changed. So, the click triggers a call to `ttt.make_move(location)`, and that in turn calls back to `view.record_change(location, old_value, new_value)` or something like that.

Chipe1:
@norvig I finished implementing the Canvas class that I had mentioned. I would like to work on showing different search algorithms on canvas and am waiting for this PR to get merged. Also, there is an example usage of TicTacToe which @SnShine can use to implement interactive TicTacToe.

PR #175, opened 2016-03-25
python_hw

natmil124:


chiragvartak:
@natmil124 Why have you committed all the changes in the `master` branch to this `aima3python2` branch? The `aima3python2` has code written in Python 2. We no longer develop it. Current development is ongoing in the `master` branch where we use Python 3 for development.

norvig:
This has gotten too big. If there is useful stuff here, it should be submitted one coherent change at a time, not as a package of changes to 50 files.

PR #178, opened 2016-03-29
fixes probability.py, removes extra notebook, update readme.md

SnShine:
Changes:
* adds truncate method instead of {0:.4f}.format().
* Merges WumpusEnvironment notebook into agents.ipynb
* Changes index of code by removing unimplemented pseudo-codes


norvig:
Good one, @SnShine , thanks.

PR #180, opened 2016-03-29
Added Details on how to view source of imported implementations

reachtarunhere:
nbviewer - http://nbviewer.jupyter.org/github/reachtarunhere/aima-python/blob/intro/intro.ipynb


norvig:
Nice -- I didn't know about %psource and the others.

PR #181, opened 2016-03-30
Cleans some parts of the codeBase

SnShine:
Changes:
* removes test for a deprecated method in test_utils.py
* obsoletes the usage of update() method defined in utils.py and changes all implementations using that method accordingly

@norvig, Is it okay to make update() obsolete?
The only use of that custom method is in `__init__` of a Class. Newly implemented Classes in the last one month don't use update() and initialises like `self.a = a`.

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -327,15 +327,15 @@ class Direction():
             d = d + "right" or d = d + Direction.R #Both do the same thing
             Note that the argument to __add__ must be a string and not a Direction object.
             Also, it (the argument) can only be right or left. '''
-    
+
     R = "right"
     L = "left"
     U = "up"
     D = "down"
-    
+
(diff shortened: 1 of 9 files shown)


norvig:
Yes, I like the idea of getting rid of update. Thanks!

I used to think that brevity was better, but now I have more experience with Python, and I realize that programmers look for a specific pattern: a sequence of `self.field = something` assignments in `__init__`, and if the code doesn't follow the pattern, it is harder to grasp,

reachtarunhere:
I wasn't expecting this reply especially since the update() trick was a part of your Python IAQ. :smile:

norvig:
The Python IAQ was before I knew what Python conventions were like ...

PR #182, opened 2016-03-30
Break out expr

norvig:
Move Expr to expr.py; add tests; no more overloading the wrong
operators for ==> and <== and the like; use ex with 3 args instead.

### DIFF
--- a/expr.py
+++ b/expr.py
@@ -1,0 +1,133 @@
+import re
+import functools
+
+@functools.total_ordering
+class Expr(object): 
+    """A mathematical expression with an operator and 1 or more arguments.
+    op is a str like '+' or 'sin'; args are either Exprs or numbers.
+    Expr('-', x) is a unary expression; Expr('+', x, 1) a binary."""
+    
+    def __init__(self, op, *args): 
+        if not isinstance(op, str):
+            raise TypeError('Expr op should be a string, not {}'.format(op))
+        for arg in args:
+            if not isinstance(arg, Expression):
+                raise TypeError('Expr arg should be an Expression, not {}'.format(arg))
+        self.op = op
+        self.args = args
+        
+    # Operator overloads
+    def __add__(self, other): return Expr('+',  self, other)
+    def __sub__(self, other): return Expr('-',  self, other)
+    def __mul__(self, other): return Expr('*',  self, other)
+    def __pow__(self, other): return Expr('**', self, other)
+    def __mod__(self, other): return Expr('%',  self, other)
+    def __and__(self, other): return Expr('&',  self, other)
+    def __xor__(self, other): return Expr('^',  self, other)
+    def __or__(self, other):  return Expr('|',  self, other)
+    def __rshift__(self, other):   return Expr('>>',  self, other)
+    def __lshift__(self, other):   return Expr('<<',  self, other)
+    def __truediv__(self, other):  return Expr('/',  self, other)
+    def __floordiv__(self, other): return Expr('//',  self, other)
+    
+    # Reverse operator overloads
+    def __radd__(self, other): return Expr('+',  other, self)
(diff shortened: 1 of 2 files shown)


Chipe1:
I suggest we keep the Expr class in logic.py and modify expr() instead of creating a new function.

norvig:
The thing about putting Expr in logic.py is that it is used elsewhere, such as in planning, where plans can be expressions. Expr as I modified it no longer has anything specifically to do with logic -- it is good for any kind of expression. So logic.py doesn't make sense; it has to be either utils.py or expr.py.

I agree that `expr` is in some ways a better function name than `ex`. But I don't like having a name that is the same as a module -- then you have to choose between `import expr` and `from expr import expr`.  So expr would be a good name if we put Expr and expr in utils.py; expr would be a bad function name if we have a sepoarate expr.py.

expr.py
```diff
@@ -0,0 +1,133 @@
...
+Expression = (Expr, int, float, complex)
+
+def arity(expression):
+    "The number of sub-expressions in this expression."
+    if isinstance(expression, Expr):
+        return len(expression.args)
+    else: # expression is a number
+        return 0
+                
+def ex(arg1, arg2=None, arg3=None):
```
SnShine: I like ex() more compared to our old method expr(). This is more flexible as we can give infix and prefix through the same function which we can't do in expr().

expr.py
```diff
@@ -0,0 +1,133 @@
...
+    """Shortcut to create an Expr, automatically defining symbols. 
+    With 1 arg, arg can be an Expr or a string denoting one.
+    With 2 args, arg1 is a unary op; with 3 args, arg2 is a binary op.
+    The other (non-operator) args can be strings or Exprs, for example:
+    >>> expr('P(x) | P(y) & Q(z)')    # 1 args
```
SnShine: need to change to ex().
ex() was used in test_ex.py so, no need to change there.

expr.py
```diff
@@ -0,0 +1,133 @@
...
+    (P(x) | (P(y) & Q(z)))
+    >>> expr('~', 'P')                # 2 args
+    ~P
+    >>> expr('P & Q', '<=>', 'Q & P') # 3 args
+    ((P & Q) <=> (Q & P))
+    """
```
SnShine: need to change to ex().
ex() was used in test_ex.py so, no need to change there.

expr.py
SnShine: I like ex() more compared to our old method expr(). This is more flexible as we can give infix and prefix through the same function which we can't do in expr().

Chipe1:
@snshine I don't see how ex() can be used for prefix. @norvig I feel ex is not a very clear name for a function. We will probably use `from expr import *` instead of `from expr import expr` because we would need the Expr class too. Or we can rename it something better than ex(), like exp() or expression ()

SnShine:
exp() is a good name.

norvig:
Let's go back to the name `expr` and move it to utils.py. 

@Chipe1 I think you're right -- there is very little advantage to prefix, because if `op` is a prefix op, then a user could easily use `expr(op + x)` or `Expr(op, x)`, depending on if `x` is a `str` or `Expression`. So it will simplify the interface if we only allow `expr(string)` or `expr(lhs, op, rhs)`.

chiragvartak:
I like `ex`; both the name and how it works. It's less confusing, more explicit. Though I am gonna miss to be able to do `expr('A & B ==> C | D')`, I think this change is necessary; `ex` makes it clearer and removes any doubt as to the precedence of operators.  
I support this Pull Request!

Chipe1:
@norvig I want to use the expr function in an implementation. Will you be merging this PR soon? I prefer having expr in utils.py and do the job of the old expr() function and your proposed ex() function. Should I write the code assuming there is such a function or should I include it in my PR.

PR #183, opened 2016-03-31
Enhances truncate() method & modifies built-in name "vars" through out the repo

SnShine:


SnShine:
@Chipe1 the main purpose of those conditions is to check if we got a matrix or a list or just a float. If we get a list, it is expected to get a list of floats. If we get a list of lists, it is expected to get floats in the lists. I don't think it's worth it to complicate things.

So, we can keep it as is or we can complicate it by checking for every value in the list if it is a float or not. And btw, if we want to do that, we need to modify most of the functions in utils.py as most of them expect some particular pattern, for example `scalar_vector_product(x, y)` which expects x to be a scalar and y to be a vector and not the other way around.

Chipe1:
I don't think a recursive code would complicate the code, instead it would simplify it. If it is a float we truncate and return. Else if it is iterable, we return the list of truncated form each element in the list. The code would be smaller, easier to understand and work for both list and matrix

norvig:
I'm going to accept this because it is a step in the right direction; thanks, @SnShine!

It would be better to break up pull requests into two parts when they address two separate things. I forgot that `vars` was a builtin, and you are right that it is good to eliminate that conflict.

As for the `truncate`, I mostly agree with @Chipe1. But a couole things. First, I think "round" is a better name than "truncate", because the .f format directive rounds, not truncates. Second, would it be ok to return rounded numbers, not strings? Or is it really important that numbers be padded with trailing zeros? If not, then I propose this, which, as @Chipe1 says, can work to any nested depth:

```
def rounder(numbers, d):
    "Round a single number, or sequence of numbers, to d decimal places."
    if isinstance(numbers, (int, float)):
        return round(numbers, d)
    else:
        constructor = type(numbers) # Can be list, set, tuple, etc.
        return constructor(rounder(n, d) for n in numbers)
```

SnShine:
Yeah, it is okay to return rounded numbers. I don't think that the numbers need to be padded with zeros. Yeah, it's better if we implement recursive method which deals with any nested depth.

@Chipe1 when I said it complicates things, I mean if we are checking for every element in the list if it is a float. I haven't read the recursion idea you proposed.

PR #189, opened 2016-04-01
Implemented LRTA* search Agent

SnShine:
Changes:
* Implementes LRTA* agent
* Adds skeleton problem classes for Maze [4.19] problem and 1-dim State Space problem [4.23].

@norvig, is it okay to have these new classes for problems? If it is alright, I will go on implementing these subproblem classes.

### DIFF
--- a/search.py
+++ b/search.py
@@ -471,6 +471,30 @@ class OnlineDFSAgent:
         self.s = current_state
         return self.a
 
+# ______________________________________________________________________________
+
+class OneDimStateSpaceProblem(Problem):
+    """ Fig. [4.23]
+    """
+    def __init__(self):
+        raise NotImplementedError
+
+    def h(self, state):
+        """
+        returns least possible cost for the given state
+        """
+        raise NotImplementedError
+
+    def c(self, s, a, s1):
+        """
+        returns a cost estimate to move from state 's' to state 's1'
+        """
+        raise NotImplementedError
+
+
+    def update_state(self, percept):
+        raise NotImplementedError
+
 
 class LRTAStarAgent:
 
--- a/search.py
+++ b/search.py
@@ -459,9 +459,52 @@ class OnlineDFSAgent:
     def update_state(self, percept):
         raise NotImplementedError
 
-def lrta_star_agent(s1):
-    "[Fig. 4.24]"
-    unimplemented()
+class LRTAStarAgent:
+
+    """Fig. [4.24]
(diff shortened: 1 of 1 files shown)


search.py
```diff
@@ -456,12 +471,77 @@ def __call__(self, percept):
...
+
+    def __init__(self, problem):
+        self.problem = problem
+        self.result = {}
+        self.H = {}
+        self.s = None
+        self.a = None
+
+    def __call__(self, s1):
+        current_state = self.update_state(s1)
```
reachtarunhere: @SnShine I guess there is a slight error here. In your implementation there is no update_state method present but it is being called.

reachtarunhere:
@SnShine I think we should try to keep the code uniform. By that I mean that my implementation of OnlineDFS agent is using percept and your implementation is using s1. We can obviously decide which one to keep and then either of us can make changes.

search.py
```diff
@@ -456,12 +471,77 @@ def __call__(self, percept):
...
+        if self.problem.goal_test(current_state):
+            self.a = None
+        else:
+            if current_state not in H:
+                H[current_state] = h(current_state)
```
reachtarunhere: I am not sure about this but I couldn't find h() defined anywhere.

search.py
```diff
@@ -456,12 +471,77 @@ def __call__(self, percept):
...
+            s = current_state
+            return a
+
+    def LRTA_cost(s, a, s1, H):
+        """
+        returns cost to move from state 's' to state 's1' plus
+        estimated cost to get to goal from s1
+        """
+        if s1 is None:
+            return(h(s))
```
reachtarunhere: @SnShine Same as above pardon me if I am wrong with this where is h() defined?

SnShine:
@reachtarunhere I am following the pseudo-code from the book where 's1' is used. I could've gone with percept but there is also 's', previous state. So, I stick with the names used in pseudo-code from 3rd edition.

SnShine:
I am a bit confused if we need the new problem subclasses, MazeProblem and StateSpace problem. If we decide not to have them, I will put unimplemented methods (which we need to override) update_state(), c(), h() in LRTA Agent.

SnShine:
If we are having those problem subclasses, h() becomes self.problem.h() and c(), update_state() accordingly.

reachtarunhere:
@SnShine I agree your code is more closely following the pseudo-code. Still I have a personal preference for percept. Once this PR gets merged I will change my code to look like yours.

PR #191, opened 2016-04-01
Implementation of Continuous World

reachtarunhere:
This is as per the discussions in #174 #160.
Essentially this is aima-demo restructured in a model view architecture.
The workflow is same as suggested by @norvig  where the model is passed to the view.

I have separated the views in a different file because they require imports specific to IPython like display. This way people who are not using jupyter can safely avoid this file.

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -516,6 +516,21 @@ class Obstacle(Thing):
 class Wall(Obstacle):
     pass
 
+
+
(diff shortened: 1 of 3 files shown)


reachtarunhere:
There is still a lot to do in terms of adding docstrings and implementing methods for collision detection. I will start working on them once this is merged.

PR #192, opened 2016-04-03
Fully implemented LRTA* agent with tests

SnShine:
Changes:
* Adds 1-dimensional state space problem as in Fig [4.23]
* Adds OnlineSearchProblem subClass
* Adds fully functional LRTA * agent
* Adds tests for LRTA* agent

Everyone, please have a look at this code and any suggestions/enhancements are welcome.

@norvig, I need a bit of clarification here:
* Firstly, this implementation is slightly different from pseudo-code. And I have added comments where it differs from pseudo-code.
* Agent takes a 'state' as input of rather than 'percept' as I am unable to find a 'percept' which specifically identifies the state in this problem space.
* I haven't used `result[state, action]` as we can/should get that from `problem.output` (can be changed to problem.result). As `result[]` is empty while initialising, it fails for some values when `LRTA_cost(self.s, b, self.result(self.s, b), self.H)` is called. So, I have used `problem.output`. Or we can initialise the `result[]` with all the possible combinations of states & actions while initialising Agent.
* There's a try: except: block in `LRTA_cost` because `H[s1]` fails for some values when `self.H[s1]` is called as it is initialised as empty table. We can avoid this by initialising `H[]` with least_possible_costs while initialising Agent.

As of now, Agent works perfectly fine with all the test cases. But, a bit deviated from the pseudo-code. I will commit changes as you wish for the above mentioned problems.

### DIFF
--- a/search.py
+++ b/search.py
@@ -459,9 +459,6 @@ class OnlineDFSAgent:
     def update_state(self, percept):
         raise NotImplementedError
 
-def lrta_star_agent(s1):
-    "[Fig. 4.24]"
-    unimplemented()
 
 # ______________________________________________________________________________
 # Genetic Algorithm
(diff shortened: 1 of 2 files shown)


SnShine:
Closing and Opening to get to pass Travis CI.

PR #193, opened 2016-04-04
Modified search.py

SnShine:
Changes:
* changed variable names like Fig[3, 4] to romania_map etc.
* removed doc tests and added unit tests to tests_.py file

### DIFF
--- a/search.py
+++ b/search.py
@@ -11,6 +11,8 @@ import random
 import sys
 import bisect
 
+infinity = float('inf')
+
 # ______________________________________________________________________________
 
 
@@ -98,7 +100,7 @@ class Node:
                 for action in problem.actions(self.state)]
 
     def child_node(self, problem, action):
-        "Fig. 3.10"
+        "[Fig. 3.10]"
         next = problem.result(self.state, action)
         return Node(next, self, action,
                     problem.path_cost(self.path_cost, self.state,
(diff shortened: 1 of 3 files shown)


reachtarunhere:
Using the Fig dictionary is a standard throughout the repo. I doubt we should be changing this as it is pretty clear in its purpose and makes the code very relatable to the book content.

SnShine:
Fig dictionary is nice, we can define things like Fig[3, 2] or Fig[3, 9] corresponding to Figure 3.2 and Figure 3.9 respectively from the book. It makes defining things easy and we can easily relate them to figures in the book. 

But, having something like `[GraphProblem('Oradea', 'Neamt', romania_map), GraphProblem('Q', 'WA', australia_map)]` is always more readable/understandable than having `[GraphProblem('Oradea', 'Neamt', Fig[3, 2]), GraphProblem('Q', 'WA', Fig[6, 1])]`. After going through book/course it is easy to visualise/recall how romania_map looks like rather than how Fig[3, 2] looks like.

@norvig, @reachtarunhere need some pointers here.

norvig:
In regards to Fig: I think there are two use cases:

- I see `Fig[3, 2]` used in the code. This has the advantage that I know where to look in the book, but the disadvantage that unless I look in the book, I don't know what `Fig[3, 2]` is. Names like `australia_map` are better for this use.

- I see "Figure 3.2" in the book, and I want to find it in the code. In this case, I won't know that it is called `australia_map`, so won't be able to search for that. I could search for "Fig[3, 2]" if I knew the code used that convention, but I could just as easily search for "Figure 3.2" if I knew that the code used the convention of putting that in a comment.

So overall, I think both use cases are best served if we get rif of the `Fig` dict and instead do things like this:

     # Figure 3.2
     australia_map = ...

SnShine:
That's nice. In this pull request I have added something like
# [Fig. 3.2]
australia_map = ...

Using 'Figure 3.2' is better than this, as you suggested. I'll update it accordingly.

PR #196, opened 2016-04-05
modified travis file to install jupyter before executing travis script

SnShine:
Removes `try: except:` block in ipyviews.py and makes the build pass by installing Jupyter on travis.


norvig:
Thanks!

PR #197, opened 2016-04-06
Fixed a typo in logic.py

Chipe1:


Chipe1:
@norvig I'm trying to implement SATPlan but I'm not sure how to convert the transitions to a cnf. Should I create a symbol for every transition and time pair and add clauses to make sure only one of the transitions is possible at a given time?

norvig:
Yes. Create lots of symbols, each one includes a time part. Then you need axioms for
- initial state
- goal
- effect of action
- exclusion: only one action can happen at a time

Here's one discussion: http://web.engr.oregonstate.edu/~afern/classes/cs533/notes/satplan.pdf

Chipe1:
@norvig What is the format of `transition` input for SATPlan algorithm. I'm assuming it is a dictionary that contains a {action->resulting state} dictionary for each state. Please correct me if I am wrong or if there is a better format.

PR #200, opened 2016-04-08
Expr with infix ops

norvig:
Add InfixOps, refactor expo and Expr

### DIFF
--- a/logic.py
+++ b/logic.py
@@ -5,10 +5,17 @@ important data types:
 
     KB            Abstract class holds a knowledge base of logical expressions
     KB_Agent      Abstract class subclasses agents.Agent
-    Expr          A logical expression
+    Expr          A logical expression, imported from utils.py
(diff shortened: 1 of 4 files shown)


logic.py
```diff
@@ -457,17 +308,18 @@ def to_cnf(s):
 
 def eliminate_implications(s):
     "Change implications into equivalent form with only &, |, and ~ as logical operators."
+    s = expr(s)
     if not s.args or is_symbol(s.op):
         return s  # Atoms are unchanged.
     args = list(map(eliminate_implications, s.args))
     a, b = args[0], args[-1]
-    if s.op == '>>' or s.op == '==>':
+    if s.op == '==>':
```
Chipe1: I think this should be >> instead of ==>
I would prefer using ==> internally instead of >> as it is more intuitive

Issue #201, opened 2016-04-08
no noqa?


norvig:
In a previous version of this code base, utils.py had support for multiple versions of Python, and did things like define the set type for old versions of Python that didn't have it. So it made sense to do "from utils import *" because you'd be importing different things for different versions of Python.

But now we're only supporting Python 3.4+, there is no such confusion, and I think it would be cleaner to replace all the "import *" with specific "from module import x, y, z" statements. That makes it easier for a reader/user to determine where a name comes from -- they need only search the current file. 

Doing this replacement will also allow us to eliminate the #noqa comments.

reachtarunhere:
I am working on this.

norvig:
Thanks!

PR #202, opened 2016-04-09
Added Explicit Imports from utils

reachtarunhere:
Fix for #201 
Long Import lines make use of brackets as per the suggestions in https://www.python.org/dev/peps/pep-0328/


reachtarunhere:
The build is failing because of an assertion error in logic.py. @Chipe1 has probably handled this in his PR.

norvig:
This is great; thanks. I think this will make it easier for readers, and if we ever want to refactor things.

PR #203, opened 2016-04-09
Fix implies operator in SAT_plan

Chipe1:
temporary fix for eliminate_implications to make the build pass


norvig:
Thanks for noticing that, @Chipe1. I should have communicated the change better, but I think this is the best way. Actually, now that we have objects for implies, etc., it might be better to use `implies.op` instead of `==>`; what do you think?

Chipe1:
@norvig Yes, but I was thinking that we bind '==>' to `implies.op` and add a new rule to 'or' `Expr` with `str`. This would enable us to do `A | '==>' | B` which looks better but makes the code harder to modify if there are any changes in the future. Should I make the mentioned change or go with replacing `==>` with `implies.op`.

norvig:
Very Interesting. So you're saying you want to have

     P |'==>'| Q

We could make that work, for sure. That's a very clever variation on
Jamitzky's trick. So we now have five proposed notations:

     P |implies| Q
     P |'==>'| Q
     implies(P, Q)
     expr('P ==> Q')
     Expr('==>', P, Q)

What I like about your idea is that it gives the user only one thing to
remember -- that implication is spelled as '==>'; the user does not also
have to remember the name 'implies'.

Another good thing about your idea is that users could do things like

     (x |'>'| y)

without having to declare "gt = InfixOp('>')" ahead of time.

To implement your idea, we would have a __or__ method on Expr that would
return a normal Expr if the rhs is an Expression, but if the rhs is a
string, it would return, let's call it PartialExpr('==>', self).  And then
PartialExpr would have a __or__ method that returns a full Expr. No __ror__
methods necessary.

Overall, I think I like your idea better, but I'm not sure what the average
user would like, so I'd like to hear if other people have opinions.

-Peter

Chipe1:
The implementation would only take one line as we already have `InfixOp` class. Instead of using pre-created instances of it(eg. `implies`, `rimplies`, `equiv`) we would create the objects directly from their strings.

norvig:
Correct -- in fact, the implementation should take a negative number of lines, since we can simplify the InfixOp class. Currently, it has to serve two purposes -- first the individual infix operators, like implies, and second, a partially filled node in the parse tree, with an op and a lhs.  If infix operators become strings, then we only need the second part.

PR #208, opened 2016-04-10
Removes obsolete Fig dictionary and modifies the codebase accordingly

SnShine:


norvig:
Good one, thanks. We should probably put the names of the Figure data structures into the table in README.md.

SnShine:
Yeah, that's a nice idea. I will do that. And @norvig, do you considered using wiki page for our repository? We can have bare essentials in our README.md and move some parts to wiki page.

PR #214, opened 2016-04-10
Implemented Grid World

reachtarunhere:
Features:

-  Ability to add labels to blocks. Labels appear in bottom right on hover.
-  Makes use of existing XYEnvironment as Model.
-  As per the Model View design decided before. Same as continuous World.
-  Ability to add custom images for representing different agents and things.
-  Uses natural syntax for creating instance no class or object name required as a parameter.
-  Walls or other things added in continuation automatically removes borders in between.

Demo: http://nbviewer.jupyter.org/github/reachtarunhere/aima-python/blob/gridtest/Untitled.ipynb

### DIFF
--- a/ipyviews.py
+++ b/ipyviews.py
@@ -1,7 +1,9 @@
 from IPython.display import HTML, display, clear_output
-
+from collections import defaultdict
 from agents import PolygonObstacle
 import time
+import json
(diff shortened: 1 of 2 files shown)


norvig:
Nice work!

PR #217, opened 2016-04-11
Implemented QLearningAgent

reachtarunhere:
I have tried to closely follow the pseudo-code. I have made use of s1, r1 instead of s_prime and r_prime as this convention is being followed by @SnShine  and others in their implementations. Later I will send a PR making changes to TD Agent and other parts I have implemented to make variable naming consistent.


norvig:
Very nice. I agree that s1, r1 are good names for s', r'.

PR #233, opened 2016-05-26
Fixing Index of Code Table

thundergolfer:
I am wanting to contribute to this repo, so I went through went to the Index of Code to find unimplemented algorithms from the book. 

I found the LRTA* Agent, though listed as unimplemented, was actually implemented. I then went through and checked the status of all apparently unimplemented methods. 

I found a few more that were actually implemented, and one that was not implemented despite being listed as complete. 

I have altered the Index of Code table to fix this.


SnShine:
Thanks for edits in README. `raise NotImplemented` in `PassiveADPAgent` class is failing doctests. Would you put that in a method __init__()?

darius:
CYK-Parse is only partly implemented; I meant to wait till it's done to add it to the list. I haven't checked the other cases, except that I remember coding Gibbs-Ask.

PR #249, opened 2016-08-25
Edited a link in CONTRIBUTING.md

rahulptel:
The link to "Pseudocode algorithm (pdf)" was pointing to https://github.com/aimacode/pseudocode/blob/master/algorithms.pdf, which was causing the 404 error. The name of the file algorithms.pdf was changed to aima3e-algorithms.pdf by @ctjoreilly in commit 80286be. Edited the link to reflect that change and point to a valid URL.


ctjoreilly:
Thanks

PR #261, opened 2016-09-14
Added test in test_planning.py

rahulptel:
Added test for spare_tire problem of planning module.


SnShine:
@rahulptel thanks for this PR.

PR #263, opened 2016-09-16
Added implementation of Three Block Tower

rahulptel:
In the precondition positive list, I am not sure whether b!=x is the best way to represent the condition. IsNot(b, x) can be used instead.

### DIFF
--- a/planning.py
+++ b/planning.py
@@ -175,3 +175,38 @@ def spare_tire():
     leave_overnight = Action(expr("LeaveOvernight"), [precond_pos, precond_neg], [effect_add, effect_rem])
 
     return PDLL(init, [remove, put_on, leave_overnight], goal_test)
+    
+def three_block_tower():
+    init = [expr('On(A, Table)'),
+            expr('On(B, Table)'),
+            expr('On(C, A)'),
+            expr('Block(A)'),
+            expr('Block(B)'),
+            expr('Block(C)'),
+            expr('Clear(B)'),
+            expr('Clear(C)')]
+
+    def goal_test(kb):
+        required = [expr('On(A, B)'), expr('On(B, C)')]
+        for q in required:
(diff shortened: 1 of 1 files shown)


SnShine:
@rahulptel I haven't worked with expr and I couldn't decide which one is better for conditions. I will leave this PR as it is until someone replies. Meanwhile you can push other algorithms.

And update the index of implemented algorithms in readme while adding new algorithms.

Chipe1:
I would suggest b == y be present in precond_neg. Also we made sure that On(X, X) never happens and On(Z, X) and Clear(X) both be possible. So we can safely remove x!=b and y!=x.
Another way would be to explicitly mention what equality means for block ie. we add Same(X, X) in the init[] and in neg_precond we append Same(b, y)

opensourceware:
As @Chipe1 said, we have to add the equality conditions in the KB through the init function (Same(x,x)). This is because expr("b != x") directly evaluates a boolean value and does not return an Expr object. Hence, the actions are failing on the PDLL formulation. Correct me if I am missing something.

Also, I feel that we don't need to put equality conditions at all as they will be taken care by the precond_check. Example: The action Move(C, C, B) will automatically fail as it doesn't (and will never) satisfy the condition On(C, C).
My suggestion might not be the right way to go about it. So more suggestions to fix the current bug are welcome. I have made a pull request for a possible solution. The commit for bug fix is at [this link](https://github.com/aimacode/aima-python/pull/265/commits/1f3b32f891cab2d8f40c48aa55ba035e78df65a2).

norvig:
@Chipe1 and @opensourceware are correct that "b != x" or "b == x" will not create Exprs; they just create booleans. An explicit Same(x, y) predicate, with a Same(x, x) axiom seems like the best approach. Either "Same" or "Equals" would be good names.

PR #268, opened 2016-10-27
Type comments

ifsdotname:
The CONTRIBUTING.md file says that at some point type annotations may be added.

I started doing that in the search.py file. I plan to continue adding them as I progress through the book and also through the AI class taught at the university where I study.

I think that they are useful, since the code is meant to accompany the book and help students understand the algorithms described better.

The comments are not in the PEP484 format for the same reason as mentioned in the contributing guide -- I do not have enough experience with it. But since the guide seems to be clearly written, I can switch to PEP484, assuming that there is someone more knowledgeable to check my pull requests.

Please let me know what you think about that.

### DIFF
--- a/search.py
+++ b/search.py
@@ -2,7 +2,14 @@
 
 The way to use this code is to subclass Problem to create a class of problems,
 then create problem instances and solve them with calls to the various search
-functions."""
+functions.
+"""
+
+from collections import defaultdict
+import math
+import random
+import sys
+import bisect
 
 from utils import (
     is_in, argmin, argmax, argmax_random_tie, probability,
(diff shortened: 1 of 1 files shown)


norvig:
I appreciate the effort, and I very much like the block comment under "Data Definition".
As for the type comments, I would rather not add them. You use your own format, and it is too much tax on the reader (makes the code longer) with not enough benefit. I think we should wait until PEP484 or something like it becomes commonplace, and then adopt the common practice. Until then, we really don't need comments that say "state" is a "State"; if there is something tricky, put it in the docstring. Can you do another pull request that just has the "Data Definition" comment?

TheCDC:
Hi, I'm new. I hope to work on this project for GSoC 2017.
Would it be worthwhile to just adopt the PEP484 as-is?

norvig:
Let's wait until the Python community as a whole comes to a conclusion about type annotations before we do anything. Closing this for now.

kaivalyar:
@norvig would it be a good idea to have something like the "Data Definition" comment for all the source files in the project? I think it will help future contributors get up to speed faster. Maybe we can have a short summary like that at the start of every file, or as a separate markdown document elsewhere in the repository?

PR #273, opened 2016-11-14
Graph plan

opensourceware:
Hi All
I have implemented a working algorithm for GraphPlan. On running spare_tire_graphplan(), the solution of the planning problem is returned. 
In the implementation, I have created separate knowledge bases for positive and negative clauses as First Order Logic KBs (class FolKB) doesn't allow non-definite clauses. Using a different set of data structures for positive and negative literals makes the algorithm simple but duplicate. We can discuss alternate ways like using a Propositional logic KB.
Also, I have created negative clauses for Persistence Actions as the "name" of the action remains the same for positive and negative clauses. Hence, I introduced a "not" in the clauses to distinguish between the persistence actions.
Thank you


opensourceware:
I hadn't squashed the commits into a single commit. sorry for the blunder. Opening a new pull request.

PR #274, opened 2016-11-14
GraphPlan Algorithm

opensourceware:
Hi All
I have implemented a working algorithm for GraphPlan. On running spare_tire_graphplan(), the solution of the planning problem is returned.
In the implementation, I have created separate knowledge bases for positive and negative clauses as First Order Logic KBs (class FolKB) doesn't allow non-definite clauses. Using a different set of data structures for positive and negative literals makes the algorithm simple but duplicate. We can discuss alternate ways like using a Propositional logic KB.
Also, I have created negative clauses for Persistence Actions as the "name" of the action remains the same for positive and negative clauses. Hence, I introduced a "not" in the clauses to distinguish between the persistence actions.
Thank you

### DIFF
--- a/planning.py
+++ b/planning.py
@@ -1,9 +1,9 @@
 """Planning (Chapters 10-11)
 """
 
+import itertools
 from utils import Expr, expr, first
 from logic import FolKB
-
 
 class PDLL:
     """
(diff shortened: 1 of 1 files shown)


norvig:
Very nice work; thank you! I'm going to accept as is, but as you mention, it would be nice to eliminate the duplication. I think that instead of using a FolKb, we can just use a set. It could be either a set of strings, or a set of Exprs. Probably strings is fine -- the idea is that GraphPlan is Propositional. Then we would need a function negate, which could be

       def negate(X): return (X[1:] if X.startswith('~') else '~' + X)

or the equivalent if using Expr instead of strings.

PR #281, opened 2017-01-16
Import turn_heading from grids.py since it is used in agents.py

orsenthil:
agents.py's XYEnvironment has turn_heading API exposed. It is internally using `turn_heading` from grids.py, but that is not imported.

This PR fixes that.

Thanks.


reachtarunhere:
Thanks for pointing out.

orsenthil:
@reachtarunhere  thanks for merging this in. Happy about my first contribution to aimacode!

PR #291, opened 2017-03-01
Formatting Fixes in Index Table

kshitij10496:
The commit 9f7f4df was the point where the build started to fail.
I found a formatting error which was causing this.
Ping @norvig


reachtarunhere:
I am fairly certain that formatting in the README is not the reason. Please check my comment on the issue. The build still seems to be failing randomly. Maybe we can set our setuptools to an older version or add six as an additional dependency to make sure it consistently passes. These would need to be tried out though.

kshitij10496:
Thanks @reachtarunhere 
This seemed a bit weird to me too but this was the only way the build could have failed from a contributor's error.
I think we can close this now as the actual reason has been encountered.
Any ideas how we can fix this random Travis build failures?

PR #294, opened 2017-03-01
Commenting Fixes

MrDupin:
Hello.

I noticed some grammar, punctuation, capitalization and formatting mistakes on some comments and I took the liberty of fixing them.

### DIFF
--- a/logic.py
+++ b/logic.py
@@ -670,7 +670,7 @@ def WalkSAT(clauses, p=0.5, max_flips=10000):
 
 class HybridWumpusAgent(agents.Agent):
 
-    "An agent for the wumpus world that does logical inference. [Figure 7.20]"""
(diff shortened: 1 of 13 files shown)


MrDupin:
Is something wrong with Travis? I only changed some commenting on the online editor. It should be able to built, but it can't.

sampadsaha5:
Nice Work. I think restarting the build would help.

logic.py
```diff
@@ -995,7 +995,7 @@ def diff(y, x):
 
 
 def simp(x):
-    "Simplify the expression x."
+    """Simplify the expression x."""
```
norvig: I'm actually not a stickler for triple quotes on short doc strings. But the PEP does suggest triple quotes, and all your other work was very helpful, so I'll agree to this.

PR #296, opened 2017-03-02
Fixing genetic_algorithm population iterator

yrevar:
Seems like a typo, results in error:
TypeError: 'int' object is not iterable.

Fixing it:
for i in len(population) -> for i in range(len(population))

### DIFF
--- a/search.py
+++ b/search.py
@@ -584,7 +584,7 @@ def genetic_algorithm(population, fitness_fn, ngen=1000, pmut=0.1):
     "[Figure 4.8]"
     for i in range(ngen):
         new_population = []
-        for i in len(population):
+        for i in range(len(population)):
             fitnesses = map(fitness_fn, population)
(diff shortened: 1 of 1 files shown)


reachtarunhere:
You are correct. Nice catch :)

reachtarunhere:
It would be really nice if you can device a test for the implementation

yrevar:
Hi @reachtarunhere, sorry I just saw your response. I am seeing that someone has already written test_genetic_algorithm(). 

In my experiment, I evaluated different randomized algorithms like hill climbing, SA, GA, and MIMIC on One-Max, Knapsack, Flip-Flop, and Continuous Peaks problems. I found that GA did quite well on Continuous Peaks problem. If we need additional tests, I can modify and add this test.

I also used AIMA search abstract class for defining all these problems. If you find it useful, I'd be happy to contribute it to this repo. You can find its snippet here: https://gist.github.com/yrevar/5f1a318de271491b6c91f14b38001fb5

Issue #287, opened 2017-02-28
Improve Test Coverage for csp.py


reachtarunhere:
The csp.py module has a lot of code but only a few tests defined in the corresponding test file. You can help by adding tests for some of the functions and methods.

kushagraagrawal:
is this issue still open?

jayantsaraf:
can i still contribute to this issue?

DeadlyHollows:
Can I still contribute to this issue?

PR #299, opened 2017-03-02
Add tests to CSP class

lucasmoura:
Add test for the following methods for the CSP class:

* assign
* unassigns
* nconflits
* actions

This MR is related to #287

### DIFF
--- a/tests/test_csp.py
+++ b/tests/test_csp.py
@@ -17,7 +17,7 @@ def test_csp_unassign():
     assignment = {var: 5}
     australia.unassign(var, assignment)
 
-    assert (var in assignment) == False
+    assert var not in assignment
 
 
 def test_csp_nconflits():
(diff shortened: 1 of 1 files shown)


tests/test_csp.py
```diff
@@ -2,6 +2,52 @@
...
+    state = {'A': '1', 'C': '2'}
+    assert map_coloring_test.actions(state) == [('B', '3')]
+
+    state = {'A': '1'}
+    assert (map_coloring_test.actions(state) == [('C', '2'), ('C', '3')] or
+            map_coloring_test.actions(state) == [('B', '2'), ('B', '3')])
+
+
 def test_backtracking_search():
     assert (backtracking_search(australia) is not None) == True
```
norvig: Instead of "assert (X is not None) == True", just do "assert X".
Here and at least one other place.

lucasmoura:
@norvig Done. I have updated the tests that were using these asserts

PR #303, opened 2017-03-03
Fix flake8 for test files

lucasmoura:
This MR adds a flake8 config file on the project root dir, with the same restrictions as the ones specified in the .travis.yml file. However, I have also added another warning to be excluded, which is the following one:

F405: name may be undefined, or defined from star imports: module

Since most of the tests are using the star import. However, if that is not a good idea, I can fix that as well.

There are some flake8 erros on test_csp.py, but when #299 gets accepted, they will no longer exist.


norvig:
Looks good, but have to fix up conflict.

lucasmoura:
@norvig Done, I have fixed the conflicts

PR #305, opened 2017-03-03
Updating Tests for learning.py

MrDupin:
Added tests for the simpler learners in learning.py (PluralityLearner, NaiveBayesLearner and NearestNeighborLearner).

Changed the input csv in test_parse_csv, as I think it would be better if we test on a non-trivial input (open to requests for change, if anyone disagrees).

Working on tests for the rest of the learners.


MrDupin:
Closed this, as I included the tests in a bigger PR.

PR #310, opened 2017-03-04
Update agents.py

clavoirr:
these aren't really the bugs but some correct standard of writing python
*functions should be in lowercase and classes title in uppercase
*convert doc string in tripple double quoted strings doc
*replace assignment with argumented assignment
*remove redundant parentheses
*argument name should be lowercase

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -42,11 +42,11 @@ import random
 import copy
 import collections
 
+
 # ______________________________________________________________________________
 
 
 class Thing(object):
-
     """This represents any physical object that can appear in an Environment.
     You subclass Thing to get the things you want.  Each thing can have a
     .__name__  slot (used for output only)."""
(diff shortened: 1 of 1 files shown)


agents.py
```diff
@@ -99,7 +98,7 @@ def can_grab(self, thing):
         return False
 
 
-def TraceAgent(agent):
+def trace_agent(agent):
```
norvig: This is a factory method, so I think it is appropriate to be UpperCase.

agents.py
```diff
@@ -108,13 +107,15 @@ def new_program(percept):
...
+
     agent.program = new_program
     return agent
 
+
 # ______________________________________________________________________________
 
 
-def TableDrivenAgentProgram(table):
+def table_driven_agent_program(table):
```
norvig: Factory

agents.py
```diff
@@ -125,55 +126,62 @@ def program(percept):
         percepts.append(percept)
         action = table.get(tuple(percepts))
         return action
+
     return program
 
 
-def RandomAgentProgram(actions):
-    "An agent that chooses an action at random, ignoring all percepts."
+def random_agent_program(actions):
```
norvig: Factory

clavoirr: surely they are but in factory methods also standard way is function name should start with lowercase

clavoirr:
surely they are but in factory methods also standard way is function name should start with lowercase

norvig:
Some good stuff here has been merged into #556.

PR #311, opened 2017-03-04
Image Mistake Fix

MrDupin:
In kNN_plot.png, I noticed a small error. Instead of "k = 6" at the bottom, it should be reading "k = 5". I amended that with this PR. It isn't a perfect fix, but I can't really do much better.


Chipe1:
I think the mistake is in the ipynb, not in the image. We are taking 6 neighbours in the image. It would be too much effort to make a new image with k = 5 without the boundary partially passing through another circle. So I suggest we keep the old image and make k = 6 in the notebook

MrDupin:
Yes, but most of literature (that I've read) on the kNN algorithm proposes an odd value for k, preferably small. That is why I think having k = 6 is not great.

norvig:
For binary classification problems, it is always better to have k be an odd value, as @MrDupin says.

PR #315, opened 2017-03-04
Update Comments in learning.py + PluralityLearner Update

MrDupin:
- Fixed some capitalization, spelling and quotation mistakes in comments

- In the PluralityLearner function, the nested function "predict" always returns the same output for a dataset, without taking into account the input ("example"). I defaulted the input as an empty list, so that we don't have to create (or find) a dummy example when we want to simply find the most popular class.

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -238,7 +238,7 @@ def PluralityLearner(dataset):
     in the training data.  Makes a baseline for comparison."""
     most_popular = mode([e[dataset.target] for e in dataset.examples])
 
-    def predict(example):
+    def predict(example=[]):
         """Always return same result: the most popular from the training set."""
         return most_popular
     return predict
--- a/learning.py
+++ b/learning.py
@@ -164,7 +164,7 @@ class DataSet:
 
 
 def parse_csv(input, delim=','):
-    """Input is a string consisting of lines, each line has comma-delimited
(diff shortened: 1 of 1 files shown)


learning.py
```diff
@@ -369,21 +368,21 @@ def plurality_value(examples):
         return DecisionLeaf(popular)
 
     def count(attr, val, examples):
-        "Count the number of examples that have attr = val."
+        """Count the number of examples that have attr = val."""
         return len(e[attr] == val for e in examples) #count(e[attr] == val for e in examples)
 
     def all_same_class(examples):
-        "Are all these examples in the same target class?"
+        """Check if all the examples are in the same target class"""
```
norvig: I think the original wording is better, because it makes it clear the answer is True or False. "Check if" does not make that as clear.

learning.py
```diff
@@ -239,8 +238,8 @@ def PluralityLearner(dataset):
     in the training data.  Makes a baseline for comparison."""
     most_popular = mode([e[dataset.target] for e in dataset.examples])
 
-    def predict(example):
-        "Always return same result: the most popular from the training set."
+    def predict(example=[]):
```
norvig: It is wrong to make example be an optional argument. The contradict for predict is that it works on an example.

PR #317, opened 2017-03-05
Fixed knn_plot.png

Chipe1:
Changed the image so that k = 5 has 5 neighbours inside the boundary instead of 6.


norvig:
Thanks, @Chipe1, and welcome back.

PR #318, opened 2017-03-05
Updated Learning Notebook (outdated)

MrDupin:
- Removed MNIST section
- Removed Numpy section
- Added PluralityLearner section (Overview, Implementation and Example)
- Updated kNN section (added Implementation and Example)


norvig:
I don't understand why you would remove MNIST? That was a very nice section.

MrDupin:
It is nice, but I don't think it does much to help understand the algorithms in learning. It is a "load-and-visualize" section, without a connection to the algorithms. I think that disconnection will confuse readers and possibly turn them away. If we keep it, we can do one of the following:

1. Use it as the main dataset. Could be done, but it takes long to classify an image.

2. Move it to the bottom of the page running the algorithms on its own, as a "real-world" example, comparing performances and such. The main dataset in that case is something simpler, like the iris one.

norvig:
I agree with your suggestion 2 -- move it to the bottom. Then we can warn students that it takes a while to run, and it doesn't stop anything below it from running.

PR #319, opened 2017-03-05
Convert Classes to Number in learning.py

MrDupin:
I couldn't see a function to convert string class names to numeric values in the DataSet class of learning.py, so I made one since it is very useful. Some learners (like the NeuralNet one) require that classes/targets be numerical.


MrDupin:
Dammit, I messed up the PR.

Apologies, I'm still learning the ins-and-outs of GitHub.

Issue #321, opened 2017-03-06
Refactoring super function call in agents.py


articuno12:
super function call in agents.py is according to python2.7 version.In python3 a simpler function call can be used. For reference https://docs.python.org/3.4/library/functions.html#super.

kushagraagrawal:
If this issue is still open, I'd like to try. Could someone guide me as to where do I start?

PR #322, opened 2017-03-06
Refactoring super function call in agents.py #321

articuno12:
Changed super(C, self).method(arg) to super().method(arg). #321


articuno12:
@norvig I ran py.test on the original repository.Tests are getting failed on it.Same tests are getting failed after my changes.
Test results when ran on current repository:
screenshot from 2017-03-06 16-49-52
Test results when the change is made :
screenshot from 2017-03-06 16-56-16

norvig:
My mistake -- thanks for pointing it out.

PR #323, opened 2017-03-06
Update learning.py mode function

lucasmoura:
In the **mode** function in learning.py, there is a problem in this the line:

`(item, count) = Counter(data).most_common(1)`

That's because the **most_common** method returns a list. Therefore, the unpacking fails. That's the reason the build is currently failing as well. This MR fixes that and adds a test for the **mode** function

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -18,7 +18,7 @@ from collections import defaultdict, Counter
 
 def mode(data): 
     """Return the most common data item. If there are ties, return any one of them."""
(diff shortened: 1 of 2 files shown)


norvig:
Silly mistake on my part -- thanks for correcting it. It was also corrected in another change.

lucasmoura:
@norvig There is just one small problem left: missing import on **utils.py** for the Counter class.

Issue #309, opened 2017-03-03
Unexpected result in mean_boolean_error


theSparta:
I have written some tests in #293 and the `py.test` fails on this particular example during the travis build.
    
     def test_mean_boolean_error():
      >       assert mean_boolean_error([1,1], [0,0]) == 1
      E       assert 0.5 == 1
      E        +  where 0.5 = mean_boolean_error([1, 1], [0, 0])
This test works fine locally as it should because `mean_boolean_error([1, 1], [0, 0])` is indeed 1 . What could be the source of this error? 

@norvig is this the expected behaviour?

Chipe1:
I did some testing on #313 and I think I know the problem. In mean(), True + True is being evaluated as True. This makes sense if we think in binary. True + True = True; '+' acts as 'or' operator in the version travis is using to test the build. I'm not sure but this is the only reason I could come up with.

norvig:
That's right @Chipe1 -- this is a known bug in statistics.mean; it should be fixed in an upcoming release of Python. Until then, we can do mean(**int**(x = y) for ...)

PR #325, opened 2017-03-06
changed mean boolean error

Chipe1:
Fixes problem in #309


Chipe1:
#323 should fix the broken build.

PR #327, opened 2017-03-06
minor edits in agents

kaivalyar:
I have fixed the bold look in the agents.ipynb notebook as on the github view of the notebook (without breaking it on the default jupyter view).

I have also standardised the docstrings in agents.py to all use three double quotes. ("""Some DocString""")

I have some suggestions regarding agents.py that I would like to work on, but I'll probably open an issue for that instead of discussing them here.


kaivalyar:
I have not edited any code, yet the build is failing. This is (possibly) already resolved by #323, which should make the checks pass here too.

PR #332, opened 2017-03-07
Added Tests for Agents.py, and fixed Direction bug discovered

kaivalyar:
I added some unit tests for "**Direction**" from **agents.py** in **tests/test_agents.py**, and in doing so discovered an error in the definition of addition. **Left + Left** was defined to be **Left**, which I corrected to **Down**.


kaivalyar:
I seem to have accidentally sent this PR from the master branch. I will resend it using the appropriate branch workflow.

PR #337, opened 2017-03-08
Fix cross validation evaluation

lucasmoura:
The cross_validation function was not working on learning.py. Running the **compare** function, which is responsible for comparing multiple algorithms against their cross validation score would generate an error. With that said, this MR does the following:

* Update cross_validation function to resemble more the book pseudocode
* Update cross_validation_wrapper to resemble more the book pseudocode
* Add tests to function **train_and_test**
* Update **print_table** function on utils.py to better handle the cross validation output.
* Add dummy size variable to some learner implementations, in order to maintain cross_validation similiarity with the book algorithm. This should be updated in the future, in order for the size parameter to be a way to better configure a learner.


lucasmoura:
Updating this MR, because #388 already fixes the bug in leave one out cross validation

lucasmoura:
Updating this MR, because #471 already updated the **print_table** function for python3

PR #348, opened 2017-03-11
Corrected Direction arithmetic in agents.py

kaivalyar:
I noticed an error in the Direction arithmetic in agents.py, and wrote a unit test to demonstrate this. Then I fixed the bug (Left + Left should result in Down, not in Left). I had earlier mistakenly sent this in #332 - this is the correct version though. Sorry for the confusion.

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -345,7 +345,7 @@ class Direction():
         elif self.direction == self.L:
             return{
                 self.R: Direction(self.U),
-                self.L: Direction(self.L),
+                self.L: Direction(self.D),
             }.get(heading, None)
         elif self.direction == self.U:
             return{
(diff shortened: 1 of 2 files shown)


norvig:
OK, it is good to fix this, but I have to say I'm not a fan of Direction. I think it is too complicated. I think all we need is a (dx, dy) tuple, and functions for add, subtract, and multiply.

kaivalyar:
@norvig Direction does seem a little too complicated for this task. It would be a good idea to replace it, however, if we are in the process of rewriting this part of the code, then why not go beyond just a tuple that just stores location coordinates, and instead stores the direction that the agent is facing too? Something like an (x, y, facing) tuple? Should I go ahead and implement this?

PR #350, opened 2017-03-11
Completed BlindDog agent examples

kaivalyar:
Updated the BlindDog agent examples in agents.ipynb and added a new class in agents.py

- Refactored the printing of the environment state in the agents.ipynb examples to be done by the environment. Also changed the `<BlindDog>` in the output to instead reflect exact agent location and action.
- Extended the BlindDog example in agents.ipynb to 2 dimensions, with a slightly modified BlindDog program to make a new "EnergeticBlindDog" for 2D.
- Added a `GraphicEnvironment` class to agents.py - to abstract the details of the GUI on jupyter so that the user can continue to focus solely on the agent algorithm without bothering with inside GUI details. Designed it such that no modifications to previous user agent program code are required, while simultaneously allowing for future updating to a better GUI without having to change user calling code.
- Demonstrated the use of the GUI environment through an example in agents.ipynb
- Added relevant markdown comments to explain all the code objectives and modifications.

### DIFF
--- a/agents.ipynb
+++ b/agents.ipynb
@@ -98,27 +98,31 @@
     "    def percept(self, agent):\n",
     "        '''prints & return a list of things that are in our agent's location'''\n",
     "        things = self.list_things_at(agent.location)\n",
-    "        print(things)\n",
     "        return things\n",
     "    \n",
     "    def execute_action(self, agent, action):\n",
     "        '''changes the state of the environment based on what the agent does.'''\n",
     "        if action == \"move down\":\n",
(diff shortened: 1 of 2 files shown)


kaivalyar:
The travis build failed due to failed imports of agent.py, as ipythonblocks wasn't installed (unlike on my local machine). I fixed this by wrapping the import statement in a try/except block. For the GraphicEnvironment to work, however, it is necessary to install ipythonblocks - should I add this to the [requirements.txt](https://github.com/aimacode/aima-python/blob/master/requirements.txt) file?

PR #354, opened 2017-03-11
Update comments and cleanup

VladKha:
- Fixed some spelling and quotation mistakes in comments
- Removed duplicate initialization in particle_filtering
- Removed explicit inheritance from object

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -159,7 +159,7 @@ class DataSet:
 
 
 def parse_csv(input, delim=','):
-    """Input is a string consisting of lines, each line has comma-delimited
+    r"""Input is a string consisting of lines, each line has comma-delimited
     fields.  Convert this into a list of lists. Blank lines are skipped.
     Fields that look like numbers are converted to numbers.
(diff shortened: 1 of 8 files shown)


MrDupin:
Hi VladKha.

In `parse_csv`, the 'r' in front of the comments is essential so that the build doesn't fail. I believe it is because of the slash ('\') included in the comment (in '\t').

If you want to remove the 'r', I believe you can escape the slash, by adding another slash in front of it. Thus you can write ('\\t').

VladKha:
Hello, MrDupin
Thanks for help
Already fixed

learning.py
```diff
@@ -413,7 +409,7 @@ def decision_list_learning(examples):
             return [(True, False)]
         t, o, examples_t = find_examples(examples)
         if not t:
-            raise Failure
+            raise Exception
```
norvig: This should be something more explicit that Exception. Maybe ValueError if there is nothing better.

PR #360, opened 2017-03-13
Added Depth First Search Visualization

ed-word:


norvig:
This is nice, but a couple things could make it better.

* I don't like using explicit color names like "red", repeatedly, in the code. The code should say what the nodes are, and then there should be exactly one place that says what color each type of node is.

* I think it should be possible to not alter the algorithms at all, but just alter the definition of Node, to achieve the same effect. I'm not sure of this, however, but it is worth exploring.

PR #361, opened 2017-03-13
corrected equivalence operator to <=> from ==>

kaivalyar:
The equivalence operator was accidentally stated as '==>' in a location. Changed it to '<=>'.


articuno12:
@kaivalyar why did you change the python version in the line number 711?

kaivalyar:
@articuno12 Jupyter automatically changes those lines based on the system configuration. I'm running python 3.4 because the aima-python repo is specifically meant to run on python 3.4.

PR #365, opened 2017-03-15
Fix: typo in Search notebook

adityasiwan:
@norvig I have fixed some of the typos in the Search.ipynb. Please have a look. :smile:

### DIFF
--- a/search.ipynb
+++ b/search.ipynb
@@ -28,9 +28,9 @@
    "source": [
     "## Review\n",
     "\n",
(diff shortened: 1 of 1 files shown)


MrDupin:
Hi Aditya.

Your fixes are correct, but I have to point out that 'optimize' and 'optimise' are both acceptable spellings of the word. The first is the American spelling while the second one is British.

I'm not sure if the repo has a preferred styling though. If it hasn't, usually in such cases edits should be made only to keep the spelling consistent across the text.

I think it's alright in this case, but keep that in mind in the future.

Thanks.

adityasiwan:
Hi @MrDupin ,

Thanks for addressing the `optimize` part. I do agree with you here that both the version of the spellings are correct. I will keep that in mind in the future.

Other than that there are some grammatical mistakes which I have updated in the PR. Please have a look into that as well.

Thanks.

MrDupin:
The rest of the changes seem good.

You will just have to wait for @norvig to merge this PR now.

PR #377, opened 2017-03-18
Depth First Search

ed-word:
Added Depth First Search Visualization
Removed explicit use of color names like "red" and added different node state variables for each corresponding color.
The code now says what the nodes are and
there is exactly one place that says what color each type of node is.

### DIFF
--- a/search.ipynb
+++ b/search.ipynb
@@ -817,26 +817,7 @@
    "metadata": {
     "collapsed": false
    },
-   "outputs": [
-    {
-     "data": {
-      "application/vnd.jupyter.widget-view+json": {
-       "model_id": "69bed6f282a24ccea4d9041a955d1293"
-      }
-     },
-     "metadata": {},
-     "output_type": "display_data"
-  },
-  {
-     "data": {
-      "application/vnd.jupyter.widget-view+json": {
-       "model_id": "9bbdb284eaab4c9b9aa21bad47797a91"
-      }
-     },
-     "metadata": {},
-     "output_type": "display_data"
-    }
-   ],
+   "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


kaivalyar:
I'm not sure why you would remove the names of the colour names:

`node_colors[node.state]` holding a value of a colour like `red` seems more intuitive than something like `final state` - am I missing something?

MrDupin:
Hi @kaivalyar. The above changes are what @norvig suggested in #360.

kaivalyar:
Got it.

However, if colours were being renamed to what they represented, wouldn't it make sense to rename the lists containing them too? `node_colors[node.state] = 'final state'` somehow doesn't seem as self explanatory as `node_type[node.state] = 'final state'`. My point is basically that it is confusing to have `node_colors` as a list, and then not put colours in it. This is what I had meant to say, @MrDupin .

MrDupin:
Ah, I see. Yes, that is a bit confusing. Maybe it needs changing.

norvig:
Unfortunately it is too difficult to resolve conflicts manually in an .ipynb.

AnishHota:
@MrDupin  There are many other searching algorithms which are still not implemented in search.ipynb. Is it not possible to implement them or they haven't been implemented yet? And what about it GUI using tkinter,should it be implemented along with the algorithm?

MrDupin:
Hi @AnishHota. You can most definitely implement the search algorithms! You don't have to also make a `tkinter` application, although it is best if the algorithm was accompanied by a graphical example in the notebooks (it's up to you though).

If you want to discuss this further, you can open an Issue to avoid commenting on old threads. Thanks!

PR #380, opened 2017-03-18
Fixed bugs in games.py

kaivalyar:
work done:

* fixed bug that would swap opponents in a game if one of them made an invalid move!
* refactored play game method to work on game objects
* displayed dynamic output per move to human players to easily choose valid actions

[travis build will continue to fail, unrelated to these changes, due to test_agents.py]

### DIFF
--- a/games.py
+++ b/games.py
@@ -136,7 +136,10 @@ def alphabeta_search(state, game, d=4, cutoff_test=None, eval_fn=None):
 
 def query_player(game, state):
     """Make a move by querying standard input."""
+    print("current state:")
     game.display(state)
+    print("available moves: {}".format(game.actions(state)))
+    print("")
     move_string = input('Your move? ')
     try:
         move = eval(move_string)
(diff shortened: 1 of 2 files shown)


games.py
```diff
@@ -255,7 +258,9 @@ def actions(self, state):
 
     def result(self, state, move):
         if move not in state.moves:
-            return state  # Illegal move has no effect
+            return GameState(to_move=('O' if state.to_move == 'X' else 'X'),
```
norvig: I'll accept this for now, but in general, this isn't right: the to_move need not be 'O' and 'X' in other games.

kaivalyar: Yes, I agree completely. I wondered the same thing. There were two reasons I chose such an implementation:

1. this `result` function is specific to the `TicTacToe` class, which inherits from `Game` - and thus won't apply to other `Game` classes. The `TicTacToe` class specifically mentions that the only valid states are 'X' and 'O' [here](https://github.com/aimacode/aima-python/blob/master/games.py#L245).
2. the same approach had already been used in the function, in line [268](https://github.com/aimacode/aima-python/blob/master/games.py#L268) below.

However, I would be happy to discuss alternative implementations. Maybe we could have a `turn` function defined in the `Game` class itself, which could then be appropriately overridden in all subclasses (including `TicTacTow`), thus effectively generalising this part of the code.

PR #386, opened 2017-03-19
Removed flake8 test for pytest directory

Chipe1:
This was added in #363 .I think it is a bad idea to put styling restrictions on tests which are not read by the readers. It is also causing the build to fail on trivialities such as missing whitespaces or blank lines.


norvig:
I think you are right, @Chipe1 -- we don't want the build to fail, and style is not so important for tests, only results are. Thanks!

Issue #393, opened 2017-03-21
Tests missing for agents.py


articuno12:
There are no available test-cases for a major part of the code.

PR #395, opened 2017-03-22
Added testcase for compare agents in agents.py

articuno12:
Added testcase for compare_agents in agents.py . #393 
The performance of ModelBasedVacuumAgent will be atleast as good as that of ReflexVacuumAgent since ModelBasedVacuumAgent can identify when it has reached the terminal state ( both locations being clean) and will perform NoOp leading to 0 performance change , whereas ReflexVacuumAgent cannot identify the terminal state and thus would keep moving , leading to worse performance as compared to ModelBasedVacuumAgent.

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -1,5 +1,6 @@
 from agents import Direction
 from agents import ReflexVacuumAgent, ModelBasedVacuumAgent, TrivialVacuumEnvironment
+from agents import compare_agents
 
 
 def test_move_forward():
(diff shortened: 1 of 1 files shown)


MrDupin:
Hello @articuno12. There are some merge conflicts now + some issues in your code. Do you have the time to fix them, or can I take care of them for you?

The main issue with the PR is grammar/English. First of all, I think the first three comments are unnecessary. Then, you use "preformance" in variable names, while it actually is "performance". There are some other grammar/spelling errors, like "as compare to", which should be "as compared to".

If you don't have the time to fix this, please let me know so that I can use your test. Thank you.

PR #396, opened 2017-03-22
Moved asserts from main code to unit tests

kaivalyar:
* Replaced two assert statements in agents.py with if based testing, and appropriate error handling.
* Wrote a unit test to verify that the edits didn't affect normal functionality.

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -297,12 +297,14 @@ class Environment:
         for it. (Shouldn't need to override this."""
         if not isinstance(thing, Thing):
             thing = Agent(thing)
-        assert thing not in self.things, "Don't add the same thing twice"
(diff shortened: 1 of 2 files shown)


articuno12:
@kaivalyar the change you have proposed in the line number 88 can have following consequences -
if a user calls the Agent while passing a wrong program , then the passed program function will be replaced to default without the user being notified which may lead to confusion on the user's side.As the purpose of assert statement there is also that user is notified of the expected type of argument if it is passed wrong , I think you should print the error to notify the user for the same and then replace it by default.

kaivalyar:
@articuno12 very true. It should be easy to fix though - it just requires a more appropriate error message. I'll add a commit that addresses this issue.

Issue #398, opened 2017-03-22
Unimplemented function in PermutationDecoder


lucasmoura:
In the PermutationDecoder class, there is a method named **score** which makes a call to **permutation_decode**. However, this function is not implemented in the project.

```python
    def score(self, code):
        """Score is product of word scores, unigram scores, and bigram scores.
        This can get very small, so we use logs and exp."""
        text = permutation_decode(self.ciphertext, code)
        logP = (sum([log(self.Pwords[word]) for word in words(text)]) +
                sum([log(self.P1[c]) for c in text]) +
                sum([log(self.P2[b]) for b in bigrams(text)]))
        return exp(logP)
```

This function was renamed in commit 516b6f26, from **decode** to **permutation_decode**. However, I could not trace even the original **decode** function, which first appear in commit 2bd2348.

Therefore, I do believe this function needs to be implemented in order for the method to work properly.

Chipe1:
Reimplemented the class in #456

Issue #304, opened 2017-03-03
Flake8 not breaking build


lucasmoura:
I have noticed in the .travis.yml file that the flake8 command does not break the build if it returns some warnings. Does that command breaks the build if only some flake8 warnings are triggered or should it not break the build at all ?

Chipe1:
@norvig Your opinion on this?
I personally think we should not impose any automated style checks. Even if we do so, styling is a subjective matter and we must first decide which rules are important enough to break the build and which rules can be relaxed.

lucasmoura:
@Chipe1 Yes, we should definitely look if that are more warnings that should be ignored by the project. Currently, we are ignoring the same rules as the ones ignored by running flake8 after a travis.yml successful build.

However, flake8 is not only used to perform style check. It is a static analysis tool as well, which can be really usefull to capture some bugs in the code. Issues such as #398 could be avoided with we allow that kind of analysis.

But again, I do agree that we should discuss what should be ignored or relaxed in order for the tool to better serve the project's needs. Also @norvig of that issue would also clarify better what these needs are.

sofmonk:
At the end, we need good documented code and that will come from imposing such checks, but what the problem is that some merged file 'A' breaks the checks because of which the updated new file B is not able to pass the build.

PR #399, opened 2017-03-22
Fix flake8 for main files

lucasmoura:
This MR fixes flake8 warnings for the main files of the project. It also:

* Excludes flake8 checking the test files, as stated in #386 
* Add flake8 checking in .travis.yml, but now it only look for the main python files in the project

This MR also adds a TODO in the text.py code related to #398 and is related also to #304

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -162,6 +162,7 @@ def rule_match(state, rules):
 
 # ______________________________________________________________________________
 
+
 loc_A, loc_B = (0, 0), (1, 0)  # The two locations for the Vacuum world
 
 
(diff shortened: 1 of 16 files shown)


kaivalyar:
I think line 19 adds flake8 testing to the travis build process again.

I am not sure if we should add this again, after it was explicitly removed in #386. While style compliance is important, I am not sure if it is necessary to have travis use flake8 to enforce it for the builds to pass.

lucasmoura:
@kaivalyar I think for the test files, it is true that flake8 was really not that usefull, since the intended users of the code would not actually read the test files. But for the main files, I do believe it is important.
It makes easier to read code, while also making it easier to contribute and review code.

Also, issues such #398 and the #337 would not be there if we allow flake8 analysis to break the build. Besides that, any unecessary check for the project can be added to the flake8 config file.

kaivalyar:
@lucasmoura yes, I agree. Style is definitely important for readability. How strictly should it be enforced, however, is something subjective we should discuss further. What to exclude using the flake config file, for example.

norvig:
I think there is no need for flake8 on the test files.

If tyhere is bad style, we can catch examples of it individually, without imposing automated tests,

agents.py
```diff
@@ -394,8 +395,9 @@ def things_near(self, location, radius=None):
         if radius is None:
             radius = self.perceptible_distance
         radius2 = radius * radius
-        return [(thing, radius2 - distance_squared(location, thing.location)) for thing in self.things
```
norvig: I actually like this indentation better than the correction

kaivalyar:
@norvig what about other non-test files though, should we run flake8 upon them for testing pep8 compatibility before sending in future PRs?

PR #401, opened 2017-03-22
Implemented Topological Sort

MrDupin:
Wrote unimplemented function `topological_sort`. To accomplish that I added an auxiliary function `build_topological` that recursively builds the topological sort of the given graph.

Finally, I updated the call to topological_sort. For the algorithm to work, we need not only the nodes of a graph but also their relations (neighbors). Originally only the node names were passed. Now it works fine.

### DIFF
--- a/csp.py
+++ b/csp.py
@@ -308,12 +308,9 @@ def tree_csp_solver(csp):
     """[Figure 6.11]"""
     assignment = {}
     root = csp.variables[0]
+    root = 'NT'
     X, parent = topological_sort(csp, root)
-    for Xj in reversed(X):
-        if(Xj == root):
-            # THIS SHOULD BE REMOVED
-            continue
-
+    for Xj in reversed(X[1:]):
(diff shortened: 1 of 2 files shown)


lucasmoura:
@MrDupin Cool work :) I just have two suggestions:

* Instead of populating the **visited** dict with a for loop, you can use a the **defaultdict** instead:

```python
from collections import defaultdict

visited = defaultdict(lambda: False)
```

* Add some tests to test_csp.py, in order to show how the method would work.

MrDupin:
Thanks for the suggestions, I made the changes.

I don't think though that Topological Sort gets stuck in cycles. When we expand a node to a neighbor, we check if the neighbor has been visited before. If it has, we simply skip it. That way we don't get stuck in a loop.

I could augment the implementation to catch cycles, but I don't think that's needed here.

lucasmoura:
@MrDupin Yes, you are right about the cycles. Since you updated the visited dict in the beginning, it should not be a problem. My mistake, sorry for that.

Issue #392, opened 2017-03-21
intersection query in relevant_pages (nlp.py) 


bionascu:
relevant_pages in the nlp.py module currently returns pages that contain the query in its entirety, but according to the textbook (p.872) it should do it "by intersecting hit lists of query words". 

My understanding is that for `query = "green apples"` and pages `p1 = "oranges and bananas"`, `p2 = "red and green apples"`, and `p3 = "plums and apples"` relevant_pages(query) should return p2 and p3, not just p2 as it currently does.

I can make this quick fix and change the corresponding test if someone confirms that it is indeed an intersection that we want.

RojerGS:
"Intersecting hit lists of query words"

Feels like the query for the word "green" gives Query1 = {p2} and the query for the word "apples" gives Query2 = {p2, p3}. Intersecting Query1 with Query2 yields {p2}, the returned, expected result. Am I missing something?

bionascu:
I'm writing intersection and the whole time I'm thinking union!! My bad.

RojerGS:
And as of now, it only returns p2? If so, that looks wrong.

bionascu:
Yes, as of now it only returns p2. But hopefully not for much longer :)

PR #402, opened 2017-03-22
intersection query for relevant_pages and corresp test

bionascu:
Updated relevant_pages() in nlp.py to do an intersection query by intersecting the hit lists of the query terms, as described in the textbook. Previously the function was (incorrectly) returning pages that contain the query in its entirety (as a phrase). 
Also updated corresponding test. This fixes issue #392

### DIFF
--- a/nlp.py
+++ b/nlp.py
@@ -293,14 +293,17 @@ def expand_pages( pages ):
     return expanded
 
 def relevant_pages(query):
-    """Relevant pages are pages that contain the query in its entireity.
-    If a page's content contains the query it is returned by the function."""
-    relevant = {}
-    print("pagesContent in function: ", pagesContent)
-    for addr, page in pagesIndex.items():
-        if query.lower() in pagesContent[addr].lower():
-            relevant[addr] = page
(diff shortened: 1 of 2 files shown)


nlp.py
```diff
@@ -293,14 +293,17 @@ def expand_pages( pages ):
...
+    """Relevant pages are pages that contain all of the query words. They are obtained by 
+    intersecting the hit lists of the query words."""
+    intersection = set()
+    query_words = query.split()
+    for query_word in query_words:
+        hit_list = set()
+        for addr in pagesIndex:
+            if query_word.lower() in pagesContent[addr].lower():
+                hit_list.add(addr)
+        intersection = hit_list if not intersection else intersection.intersection(hit_list)
```
Chipe1: This would lead to problems when the intersection is an empty set. When the set becomes empty it is replaced by hit_list instead of staying empty. I've fixed this and the merge conflict in #509 .

norvig:
Due to @Chipe1 comment about #509, I'm going to close this for now ... could bring back the test later.

PR #403, opened 2017-03-23
Updated implementation of FIFOQueue 

articuno12:
1) Updated implementation of FIFOQueue in Utils.py using the inbuilt python library queue.Queue as stated in TODO. 
2) Added appropriate error handling which was missing earlier.
3) Wrote a detailed test case for each method of the data structure which uses random numbers so as to cover wider range of possible test data . 
<br>
About Test Case : 

- Generated an array of length 100 of random numbers to be used for testing.
- Used 2 pointers to indicate section of the array that should be present inside the queue.
- At each step , randomly with probability 0.5 either add an element to the queue ( if queue is not full) or remove an element from the queue (if queue is not empty) .
- Whenever an element is removed from the queue it is checked that the element returned is same as that pointed by the back pointer .
- Also at each step  it is checked if the length of the queue returned is equal to the difference between the two headers , indicating the length of the array that should be inside the queue .
- To test extend method , 2 test arrays each of length 50 are created . Both of them are then added to the queue using the method extend .
- Each element is then removed from the queue and compared with the elements in the test array.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,6 +1,6 @@
 import pytest
 from utils import *  # noqa
-
+import random
 
 def test_removeall_list():
     assert removeall(4, []) == []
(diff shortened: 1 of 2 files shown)


lucasmoura:
@articuno12 Nice, just a couple of suggestions:

* You check before adding an element to the queue if it is full, however, by default, the maxsize variable in the Queue class is set to 0, which means the size of the Queue will be infinite. Therefore, with that variable cannot be modified when creating the Queue, checking if the queue is full much not serve much purpose. Maybe adding the maxsize variable in the class constructor and give it a default value would be a better approach.

* Although using Queue was stated in the TODO, maybe there exists a more simple approach. Queue is mainly used for thread programming, specially for thread communication. Using a simpler structure, such as deque, could allow the same implementation and would not require, for example, the implementation of the __contains__ method as it is, since now you need to parse the queue manually to verify if it has a given value.

articuno12:
@lucasmoura Thanks for the suggestion ! 
I will include your first suggestion . 
For the second one , I did what was written in TODO . After your suggestion , I compared the two and found out that indeed deque is a better approach for implementing this as it contains methods like 'contains', 'len', 'extend', 'append' which will reduce implementation. Also , deque will be more efficient . I ran the following test 

```
import time
import queue
import collections

q = collections.deque()
t0 = time.clock()
for i in range(100000):
    q.append(i)
for i in range(100000):
    q.popleft()
print ('deque', time.clock() - t0)

q = queue.Queue()
t0 = time.clock()
for i in range(100000):
    q.put(i)
for i in range(100000):
    q.get()
print ('Queue', time.clock() - t0)
```
Results : 
```
deque 0.023917
Queue 0.445446
```
So , @norvig Should I go ahead with deque implementation?

norvig:
Yes, deque is certainly the best choice (it was not available when the previous version was written).

PR #406, opened 2017-03-23
Use lru_cache decorator on memoize function

lucasmoura:
This MR use the lru_cache decorator for caching on the memoize function. 

I have kept the slot functionality, because I think it has a differente purpose than applying the lru_cache. However, if that is not the case, I can replace it with the lru_cache decorator as well.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -270,13 +270,10 @@ except ImportError:
 # Misc Functions
 
 
-# TODO: Use functools.lru_cache memoization decorator
-
-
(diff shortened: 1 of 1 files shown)


norvig:
This change is going in the right direction, so I will accept it, but probably what we should do is replace `memoize` with `lru_cache`, except if we really need the `slot` caching (I'm not sure), then add a `attr_cache` function that does just that.

PR #407, opened 2017-03-23
Update test_learning

MrDupin:
Learners currently are binary classifiers, so I used `remove_examples` to remove extra classes from the dataset.


MrDupin:
Merged edits with #410.

PR #410, opened 2017-03-24
Bug Fixing in DataSet + Test Updates

MrDupin:
In `learning.py`, the DataSet class had some bugs and issues that I fixed:

* `exclude` wasn't working. When mapping a function in Python 3, it returns just a pointer. I converted the mapping to a list.

* `remove_examples` wasn't removing the example values from the `values` variable. I added the feature.

Also, I added a test for `exclude` in `test_learning.py` (+ fixed a minor spacing inconsistency) and since learners currently are binary classifiers, I used `remove_examples` to remove extra classes from the dataset.

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -116,7 +116,7 @@ class DataSet:
             self.inputs = [a for a in self.attrs
                            if a != self.target and a not in exclude]
         if not self.values:
-            self.values = list(map(unique, zip(*self.examples)))
+            self.update_values()
         self.check_me()
 
     def check_me(self):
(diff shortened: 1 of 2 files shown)


lucasmoura:
@MrDupin Although the values should be updated when we remove a example, I don't think it is good to replicate code, since the code on line 119 is the same as the one in 170. Maybe create an auxiliary method to update the values would be a better approach:

```python
def update_values(self):
   self.values = list(map(unique, zip(*self.examples)))
```

And then call this method on both line 119 and 170.

MrDupin:
Good catch, thanks.

Issue #409, opened 2017-03-24
LinearLearner Issues


MrDupin:
I'm having some issues with the `LinearLearner` function in learning.py and I would like some help.

I noticed that the function does not seem to have pseudocode, even though in the script it is a standalone algorithm. At first it had some bugs and wouldn't execute, but when I fixed them (#408) it returned a result that confused me.

It does not use an activation function, so it quite probably does regression of sorts. But the given result is way off. This is what I did:

I tested the function on the Iris dataset. First, I converted classes names to numbers. So the targets are 0,1,2. Then I run the `LinearLearner` on the modified dataset. Every time it returns very large or very small numbers (eg. 1.00974582882851e+128). Even though technically I am asking for classification, it shouldn't produce numbers that off, since classes are converted to numbers in [0, 2].

About the same results are given when I remove one class to make this a binary problem. I even tested this with an activation function, but the numbers were so big that an OverflowError was raised.

Is there something I misunderstood about the function? If it is indeed wrong, since we don't currently have pseudocode, how should we act? Delete it? Try and make it work? Write a warning?

---
PS: The code I used to test the algorithm is the following:

<pre>
def test_linear_learner():
    iris = DataSet(name="iris")

    classes = ["setosa","versicolor","virginica"]
    iris.classes_to_numbers()

    learner = LinearLearner(iris)
    assert learner([5,3,1,0.1]) in range(len(classes))
</pre>

lucasmoura:
@MrDupin I believe the gradient descent part is incorrect. We are using batch gradient descent, and therefore, the gradient value should be divided by the number of examples. Also, given that the error is:

```python
err = t - y
```

I believe that the gradient formula should be:

```python
w[i] += w[i] + learning_rate * (dotproduct(err, X_col[i]) / len(examples))
```

I will send a MR with this fix, but maybe there is something more that I could not see.

MrDupin:
Discussion continued on #414.

PR #414, opened 2017-03-24
Fix gradient descent for LinearLearning

lucasmoura:
There is a bug in the gradient descent in LinearLearner, as discussed in #409. This MR deals with the problem.

The main problem is the use of batch gradient descent without dividing the value by the number of examples.

Also, this MR depends on the bug fixes for LinearLearner provided by #408

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -635,6 +635,7 @@ def LinearLearner(dataset, learning_rate=0.01, epochs=100):
     idx_i = dataset.inputs
     idx_t = dataset.target  # As of now, dataset.target gives only one index.
     examples = dataset.examples
+    num_examples = len(examples)
 
     # X transpose
     X_col = [dataset.values[i] for i in idx_i]  # vertical columns of X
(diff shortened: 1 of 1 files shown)


MrDupin:
This certainly nudges the implementation in the right direction, removing the issue of overflow errors, but still I don't think this is complete. The problem is I'm not sure if the learner is supposed to accomplish classification or regression.

I have not heard of a Linear Learner algorithm, so I assume this is a general linear learning algorithm. I believe this type of algorithms works by drawing a straight line across points in a space. This essentially classifies the points into two classes (above/below line).

If this is indeed a classifier, we should pass the result through an activation function when making a prediction. I suggest we change the `predict` function to something like this:

<pre>
def predict(example):
    x = [1] + example
    pred = sigmoid(dotproduct(w, x))
    return 1 if pred >= 0.5 else 0
</pre>

The above will try and classify the given item into one of two classes. After I tested this a little, I found that it actually classifies items with a pretty good rate (of the 5 items I tested, it got right 4).

PR #415, opened 2017-03-25
added fol_fc_ask to logic.py ;  update README.md 

sofmonk:
1. updated **logic.py** :added **fol_fc_ask from fig 9.3** (it showed implmemented in README.md but was NotImplemented) 

2. updated **readme.md** : added unimplemented algorithms
10.1 Air-Cargo-problem
10.2 Spare-Tire-Problem
10.3 Three-Block-Tower
10.7 Cake-Problem
10.9 Graphplan
11.10 Doubles-tennis

3. updated **planning.py** : minor pep8 fixes


Chipe1:
Could you add a test case for fol_fc_ask. The variable 'new' is clearly defined inside the loop while whereas we are already checking if it is None before entering the loop.

sofmonk:
I am working on the test files.
You can think of it as a 'do-while loop', after each iteration of the loop, it checks for emptiness of 'new' and again loops.

Chipe1:
Also, I'm not sure whether random.KB exists. I tried running that and it throws an error. Do we need to import somthing to make it work?

sofmonk:
Could you please share the test case you used?

Chipe1:
First I added new  = [] before starting the while loop and changed 'for rule in KB' to 'for rule in KB.clauses'

PR #417, opened 2017-03-25
Update learning.py

Agent-Pandit:
converted method sample(self) to propety and removed the call in the statement return self.sampler


Chipe1:
search.ipynb has an unnecessarily long output which makes it 2.45 MB!
It is better to have the images in the repo itself rather than fetching them from external sources.
You might have to change the newline character on the editor you used. It replaced all '\n' with '\r\n'. Notepad on windows does that. Or you can use notepad++ or any other non-windows editor/IDE like emacs

PR #418, opened 2017-03-25
Removed errors to make the build pass

articuno12:
Removed flake8 errors causing build to fail.

Issue #416, opened 2017-03-25
Build Failing


articuno12:
Build is failing as test_text_models in test_text.py is not getting passed.
Result of running tests on the current repository : 
```
tests/test_agents.py::test_move_forward PASSED
tests/test_agents.py::test_add PASSED
tests/test_agents.py::test_ReflexVacuumAgent PASSED
tests/test_agents.py::test_ModelBasedVacuumAgent PASSED
tests/test_csp.py::test_csp_assign PASSED
tests/test_csp.py::test_csp_unassign PASSED
tests/test_csp.py::test_csp_nconflits PASSED
tests/test_csp.py::test_csp_actions PASSED
tests/test_csp.py::test_csp_result PASSED
tests/test_csp.py::test_csp_goal_test PASSED
tests/test_csp.py::test_csp_support_pruning PASSED
tests/test_csp.py::test_csp_suppose PASSED
tests/test_csp.py::test_csp_prune PASSED
tests/test_csp.py::test_csp_choices PASSED
tests/test_csp.py::test_csp_infer_assignement PASSED
tests/test_csp.py::test_csp_restore PASSED
tests/test_csp.py::test_csp_conflicted_vars PASSED
tests/test_csp.py::test_revise PASSED
tests/test_csp.py::test_AC3 PASSED
tests/test_csp.py::test_first_unassigned_variable PASSED
tests/test_csp.py::test_num_legal_values PASSED
tests/test_csp.py::test_mrv PASSED
tests/test_csp.py::test_backtracking_search PASSED
tests/test_csp.py::test_universal_dict PASSED
tests/test_csp.py::test_parse_neighbours PASSED
tests/test_csp.py::test_topological_sort PASSED
tests/test_games.py::test_minimax_decision PASSED
tests/test_games.py::test_alphabeta_full_search PASSED
tests/test_games.py::test_random_tests PASSED
tests/test_grid.py::test_distance PASSED
tests/test_grid.py::test_distance_squared PASSED
tests/test_grid.py::test_vector_clip PASSED
tests/test_learning.py::test_exclude PASSED
tests/test_learning.py::test_parse_csv PASSED
tests/test_learning.py::test_weighted_mode PASSED
tests/test_learning.py::test_weighted_replicate PASSED
tests/test_learning.py::test_plurality_learner PASSED
tests/test_learning.py::test_naive_bayes PASSED
tests/test_learning.py::test_k_nearest_neighbors PASSED
[...]

PR #420, opened 2017-03-25
Fixed bug with for loop

RojerGS:
Limit of the foor loop was missing the last ngram of the text and hence missing a test that specifically addressed that.

### DIFF
--- a/text.py
+++ b/text.py
@@ -80,7 +80,7 @@ class NgramTextModel(CountingProbDist):
 
 class NgramCharModel(NgramTextModel):
     def add_empty(self, words, n):
-        return  ' ' * (n - 1) + words
+        return ' ' * (n - 1) + words
 
     def add_sequence(self, words):
         for word in words:
--- a/text.py
+++ b/text.py
@@ -60,7 +60,7 @@ class NgramTextModel(CountingProbDist):
         n = self.n
         words = self.add_empty(words, n)
 
-        for i in range(len(words) - n):
+        for i in range(len(words) - n + 1):
             self.add(tuple(words[i:i + n]))
 
     def samples(self, nwords):


RojerGS:
This addresses the same issue as #412 that, for some reason, was reverted with #413 . 412 also introduced new tests that addressed the issue covered by the bugfix. When the 413 reverted the bugfix, those new tests started failing.

articuno12:
@RojerGS I already fixed flake8 error in #418

RojerGS:
@articuno12 I have seen it. I only addressed this one because I hadn't noticed before.

articuno12:
@RojerGS A small advice, in open source, you should read what changes others have done before committing yours !

RojerGS:
@articuno12 you are most right, my apologies.

RojerGS:
@Chipe1 after your #430 I no longer know if I should fix the flake8 warnings like you asked in #416 
Also, as @articuno12 notes, the work was already started by him/her. Should I do it nonetheless?

Chipe1:
@RojerGS #430 only removes a redundant recheck. Flake8 is still necessary for the build to pass. I think it would be better if there was a single PR that fixes all flake8 errors. Different PRs fixing the problem partially would be hard to track. The errors are very easy to fix and there is no point in breaking them into smaller components. I suggest you copy the changes which @articuno12 made and fix the other errors so that there would not be any merge conflicts. Or we can wait till @articuno12 fixes all the flake8 related problems in his PR

articuno12:
@Chipe1 I will fix them all in my PR.

Chipe1:
This is not the currect thread to discuss this but #415 adds fol_fc_ask which seems to be incomplete and causes flake8 errors. @articuno12 would either have to remove it, or @sofmonk needs to update the implementation.

articuno12:
@Chipe1 For now I have changed the lines causing flake8 errors in fol_fc_ask, this removes flake8 errors but doesn't guarantee fol_fc_ask to work.

PR #422, opened 2017-03-26
Implemented 11.1 Job-Shop-Problem

kaivalyar:
11.1: Job-Shop-Problem-With-Resources (previously unimplemented)

* Created HLA class to model chapter 11 HLAs in planning.py as opposed to the chapter 10 Actions (by inheriting from and extending the Action class)
* Similarly, created the (real-world) Problem class to model problems with resources and ordering constraints from chapter 11
* Wrote the `job_shop_problem` function that builds and returns the problem as specified in figure 11.1
* Made the file pep8 compatible using flake8, wrote docstrings, and added a unit-test for all of the above.

I am hoping to now use the HLA and Problem classes to implement Refinement, Hierarchical Search, and Angelic Search. I sent this PR despite not having finished those yet just to check if my approach for modelling HLA and Problem were okay. I will update the code with the remaining chapter 11 implementations soon if these are okay.

(As mentioned in #274 using FolKB may be unnecessary. However, since HLA inherits from Action, corresponding changes to the Action implementation should already reflect automatically. If there are any other similar design issues that may require changes, I would be happy to modify this PR accordingly.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Here is a table of algorithms, the figure, name of the code in the book and in t
 | 10.7    | Cake-Problem       |   `have_cake_and_eat_cake_too`       |[`planning.py`][planning]|
 | 10.9    | Graphplan          |    `GraphPlan`      |[`planning.py`][planning]|
 | 10.13   | Partial-Order-Planner |          |
-| 11.1    | Job-Shop-Problem-With-Resources |          |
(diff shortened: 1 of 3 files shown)


kaivalyar:
The build failure is a result of #413 getting merged, as stated in #416.

kaivalyar:
This PR has developed merge conflicts, so closing it and creating a new one.

PR #427, opened 2017-03-27
added examples to notebooks LOGIC & PLANNING

sofmonk:
1. added examples to **logic.ipynb**:

- PropKB
- tt_entail
- tt_check_all
- dpll_satisfiable
- WalkSAT
- SATplan
- unify

2. added examples to **planning.ipynb**:

- factored state representation
- PDLL
- Actions
- Air_cargo_problem
- DWR (dock worker robot) problem
- spare_tire_problem
- blocks_world

_Sir, I found some examples/diagrams from the other websites good, but I didn't include them as I was skeptical if we can add content from other websites. Please let me know so that I can add them?_


MrDupin:
I will paste some of my feedback from [my previous comment](https://github.com/aimacode/aima-python/pull/424#issuecomment-289293067) so that it doesn't get lost.

* Spacing issues. You have inconsistent spacing in the text. For example, you write "precondition(positive and negative)" and "same way as before :". In the first case you need a space before the parenthesis, while in the second case you need to remove the space before the colon.

* Capitalization issues. When starting a new sentence, you should be using capital letters even if before that you had a code example. It still is a sentence + you break capitalization consistency.

* Formatting/Styling. Most of the Notebooks utilize formatting to help ease the reader in. Italics, bold, etc. help a lot and look nice. A lot of the time I was not sure if what I was reading was an actual word or a function/class name. This can be easily fixed by styling function/variable/etc names as code. You can do that by wrapping the name in left-leaning quotations (the quotation usually found under the 'tilde' key -> ``). It produces more or less the following effect: `code`. You can use italics and bold font in similar instances too.

---

Like before, the notebooks you provided are lacking in description. While you did improve this since last time, there is still a lot missing. The majority of the examples you provide are not adequately explained and leave a lot up to the reader. I'm not saying explain every little detail, but you need to write a bit more.

As an example take the following:

<pre>
KB.tell(A & C)
KB.ask(A) == KB.ask(C) == {}
</pre>

You write the above code and the explanation you write is this: "adds sentence and checks if KB entails the query."

This brings up a lot of questions: Is "A & C" the sentence? What does "KB.ask(A)" do? Why does it return an empty dictionary? What exactly *is* "A"?

All these questions can be easily answered by a couple of sentences.

---

This is definitely an improvement from last time and you are on the right track, but in my opinion this still needs work before it gets merged.

Again, this is ultimately up to @norvig, but I would take quality and thoroughness over quantity.

MrDupin:
I like these additions. Very nice work.

Did you make the images yourself?

sofmonk:
@MrDupin Thank you, your suggestions helped out. 
I got those images from other websites. I am still doubtful if we can use other web-site's content?

MrDupin:
That depends, but maybe not. If you share where you found them I can look into this more.

Chipe1:
@sofmonk That depends. In most cases you are not allowed to reuse the images without permission. It would be better if we used images in the textbook or make them on our own.
I would recommend opening an issue and posting what images are needed. Me and other contributors can create the required images and upload them.

norvig:
This is good, I appreciate the work on this! But we need to resolve some issues -- the branch has conflicts, the rights to the photos have to be figured out (we need to get permission to the author (or the author has to state a license agreement like creative commons) and we have to give credit to the original author. And there are a few comments from MrDupin should be addressed.

PR #429, opened 2017-03-28
Viterbi Calculation Reusability + Comment Fixes

MrDupin:
In `viterbi_segment`, the calculation `P[w] * best[i - len(w)]` is used twice in a row, so I stored it into a variable (`curr_score`) and used that instead.

Also made quotation and spacing fixes to the comments.

### DIFF
--- a/text.py
+++ b/text.py
@@ -18,11 +18,11 @@ import os
 class UnigramTextModel(CountingProbDist):
 
     """This is a discrete probability distribution over words, so you
-    can add, sample, or get P[word], just like with CountingProbDist.  You can
-    also generate a random text n words long with P.samples(n)"""
+    can add, sample, or get P[word], just like with CountingProbDist. You can
+    also generate a random text n words long with P.samples(n)."""
 
     def samples(self, n):
-        "Return a string of n words, random according to the model."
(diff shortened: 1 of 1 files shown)


text.py
```diff
@@ -97,12 +97,13 @@ def viterbi_segment(text, P):
...
     words = [''] + list(text)
     best = [1.0] + [0.0] * n
-    # Fill in the vectors best, words via dynamic programming
+    # Fill in the vectors best words via dynamic programming
     for i in range(n+1):
         for j in range(0, i):
             w = text[j:i]
-            if P[w] * best[i - len(w)] >= best[i]:
-                best[i] = P[w] * best[i - len(w)]
+            curr_score = P[w] * best[i - len(w)]
```
MrDupin: I have seen `curr_something` (or `currSomething` etc.) pop up a lot around, so I assume most people will get it. The problem with `p_score` is that `p` doesn't give enough information, which is not the same case as `curr_score`.

I personally feel it's fine, but I don't feel strongly about it. If more have the same issue, I will change it.

Chipe1: You are right. I'm sure most won't have an issue.

MrDupin:
Branch was deleted, will make a new PR.

PR #430, opened 2017-03-28
Removed post-build flake8 checks

Chipe1:
#399 makes flake8 compulsory for the build to pass. We no longer need to run flake8 after the build is successful.


lucasmoura:
@Chipe1 Yes, we don't need flake8 to run after a successful build, but I don't think that just running flake8 after a successful build will serve any purpose, since even if it display warning, it will never affect the build status.

PR #432, opened 2017-03-28
Implementation: Tree CSP Solver

MrDupin:
Completed the implementation of `tree_csp_solver` and added tests. In detail:

* Implemented `make_arc_consistent`. For a node *n* and its parent *p*, it removes the possible values of *p* which do not have a consistent assignment with any possible value of *n*.
* Added new function, `assign`. It assigns a value to a node consistent to the assignment of the node's parent.
* Added new function, `node_domains` which converts `UniversalDict` to a built-in dictionary. This way we can play around more with the dictionaries.
* Added test for the algorithm. The test uses the example from the book (Australia with a node removed).

Note: I had to add two new lines to the algorithm for initialization purposes. If that's an issue, I can work them into the called functions, but it will be ugly.

### DIFF
--- a/csp.py
+++ b/csp.py
@@ -85,7 +85,7 @@ class CSP(search.Problem):
         # Subclasses can print in a prettier way, or display with a GUI
         print('CSP:', self, 'with assignment:', assignment)
 
-    # These methods are for the tree- and graph-search interface:
+    # These methods are for the tree and graph-search interface:
 
     def actions(self, state):
         """Return a list of applicable actions: nonconflicting
@@ -152,6 +152,13 @@ class CSP(search.Problem):
         """Return a list of variables in current assignment that are in conflict"""
         return [var for var in self.variables
                 if self.nconflicts(var, current[var], current) > 0]
+
+    # This is for tree_csp_solver
+
+    def node_domains(self):
+        """To be used only if self.domains is a UniversalDict.
+        Converts self.domains to a built-in dictionary."""
+        return defaultdict(lambda: self.domains.value)
 
 # ______________________________________________________________________________
 # Constraint Propagation with AC-3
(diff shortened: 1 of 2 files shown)


lucasmoura:
@MrDupin Nice work, but just some suggestions:

* Instead of creating the **node_domains** method, wouldn't it be better to just call the **support_prunning** method before running the tree_csp_solver ? That way, inside the
**make_arc_consistent** method, you can just call the prune method with the invalid variables, you will not need to create a new domain for a given variable.

* I don't think the **break** variable is usefull in the code, since in your current implementation you could just append the variables if the constraint returns True.

With that said, my ideia  for the code was something like this:

```python
def make_arc_consistent(Xj, Xk, csp):
    """Make arc between parent (Xj) and child (Xk) consistent under the csp's constraints,
    by removing the possible values of Xj that cause inconsistencies."""
    for val1 in csp.domains[Xj]:
        for val2 in csp.domains[Xk]:
            if not csp.constraints(Xj, val1, Xk, val2):
                csp.prune(Xj, val1)

    return csp.curr_domains[Xj]
```

But again, this will require calling the **support_prunning** method. But this is just an idea.

* I don't know if **assign** is a good name for the function you are creating. I understand your intention, but when I see a method called **assign**, I assume it will have a different behavior. Specially since CSP has an assign method too. Maybe something like **get_first_valid_value**  or something like that will make it clearer what the function is doing. But even my suggestion can be improved.

* Maybe it will be good to add a test on a example that does not have a solution, just to clarify what will happen in that case.

MrDupin:
@lucasmoura: Thanks for the feedback.

* I will take a look at `support_pruning` since it seems more general than `node_domains`. With `node_domains` I wanted to write something specific for `UniversalDict`, but I can see how that can be restricting.

* The `break` is not necessary, but is useful. We only need to find one possible value that satisfies the constraint in order to keep `val1`, so I break when I find one.

* I would have used `prune`, but unfortunately we cannot use it the way you mentioned. We need to check all the possible values `val2` before making a decision on removing/keeping `val1`, so we cannot prune first time the constraint is not satisfied. I will look into using `prune` after the inner loop though, since then it will work fine.

* `assign` is necessary, yes. `make_arc_consistent` makes sure that there is a possible assignment for any value of the parent, but since we don't know in advance what the value of the parent will be, we cannot assign a value to the child. With `assign` we pick the first value in a node that satisfies the constraints given its parent's value. I will change the name though, since I somehow missed that `CSP` had a same name function. Good catch.

* I am looking into adding a failing example, but unfortunately constraints are now a bit limiting. Since input is a tree, we cannot use map coloring constraints since for any non-trivial problem (aka with more than 1 color) there will be a solution. I might in the end put such an example in just to have a test, but I want to work on something more substantial first.

* What do you mean *two commits with the same name*? The two updates on the `test_csp`?

MrDupin:
Accidentally deleted branch, will now make new PR. Sorry about that.

PR #433, opened 2017-03-28
Notebook: Tree CSP Solver

MrDupin:
Added the "Tree CSP Solver" algorithm to `csp.ipynb`. Includes an overview of the algorithm and the implementation + an example.

For the code in the Notebook to execute, #434 needs to be merged as well.

### DIFF
--- a/csp.ipynb
+++ b/csp.ipynb
@@ -15,7 +15,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 1,
    "metadata": {
     "collapsed": true,
     "deletable": true,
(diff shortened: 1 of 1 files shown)


MrDupin:
For the code in this to execute properly, #434 needs to be merged as well.

sofmonk:
The notebook has missing outputs. Kindly, run the remaining cells.

MrDupin:
I left everything outside my commit as it was, so the outputs were probably supposed to be that way. I'm not sure if I should be changing that here.

If yes, I will, but it is not urgent since the users can run them themselves. You can create a new issue asking about it.

PS: I ran the cells and they do not seem to be displaying important outputs. Some outputs are very long, some have warnings and others are numbers that are not mentioned in the rest of the Notebook. I think the outputs are supposed to be hidden.

sofmonk:
in response to your PS: The outputs include visualizations for Graph coloring problem and N-Queens problem.

MrDupin:
Yes, but unfortunately they throw huge red warning squares (at least for me). You can take care of it if you want, but I don't have the time right now, I'm sorry.

Chipe1:
According the discussion in #156 the notebooks by default must have the outputs stored. But seeing as the repository is currently in a bad shape with the build failing and many crucial commits still not merged it would be better if  we wait till this PR is merged before making further changes to prevent any merge conflicts. Git and Jupyter don't work too well together, sadly.

norvig:
The notebook should have the outputs stored. This is difficult when the output includes an interactive component. I'm not sure of the best way to handle this. I will accept this request, and we should move forward, and continue looking for a better way to handle notebooks.

PR #434, opened 2017-03-29
Implementation: Tree CSP Solver

MrDupin:
Completed the implementation of `tree_csp_solver` and added a test. In detail:

* Implemented `make_arc_consistent`. For a node *n* and its parent *p*, it removes the possible values of *p* which do not have a consistent assignment with any possible value of *n*.
* Added new function, `assign_value`. It assigns a value to a node consistent to the assignment of the node's parent.
* Added test for the algorithm. The test uses the example from the book (Australia with a node, 'SA' removed).

Note: I had to add two new lines to the algorithm for initialization purposes. If that's an issue, I can work them into the called functions, but it will be ugly.

---

PS: Props to @lucasmoura for his feedback [here](https://github.com/aimacode/aima-python/pull/432#issuecomment-289943075). Also, I accidentally deleted the previous branch, so I had to make a new PR. Apologies. Previous PR is #432.

### DIFF
--- a/csp.py
+++ b/csp.py
@@ -85,7 +85,7 @@ class CSP(search.Problem):
         # Subclasses can print in a prettier way, or display with a GUI
         print('CSP:', self, 'with assignment:', assignment)
 
-    # These methods are for the tree- and graph-search interface:
+    # These methods are for the tree and graph-search interface:
 
     def actions(self, state):
         """Return a list of applicable actions: nonconflicting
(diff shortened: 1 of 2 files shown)


sofmonk:
The notebook has missing outputs. You will have to run the remaining cells, they show no output.

norvig:
Thanks for this, and thanks @lucasmoura for some nice comments.

PR #437, opened 2017-03-29
Refactor backpropagation

lucasmoura:
Just some minor refactoring in the BackPropagationLearner, such as:

* Create a function to initialize the random weights
* Create a function to calculate the sigmoid value

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -3,7 +3,8 @@
 from utils import (
     removeall, unique, product, mode, argmax, argmax_random_tie, isclose, gaussian,
     dotproduct, vector_add, scalar_vector_product, weighted_sample_with_replacement,
-    weighted_sampler, num_or_str, normalize, clip, sigmoid, print_table, DataFile
+    weighted_sampler, num_or_str, normalize, clip, sigmoid, print_table,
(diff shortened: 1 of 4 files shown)


MrDupin:
Nice work. A couple of suggestions:

* In `utils.py` we have a function for the sigmoid. It takes as input a number and returns a number. It would be better if we had the same for its derivative. I suggest you write `sigmoid_derivative`, with arguments a number instead of a node. Then you could call this function with `sigmoid_derivative(o_nodes[i].value)`. Also, as a name, `sigmoid_derivative` is cleaner than `sigmoid_derivative_value`, since we also have `sigmoid` and not `sigmoid_value`.

* For the random weight function, in my opinion a better name would simply be `random_weights`. Currently the call to the function is too long (almost as long as the code it replaces). If `random_weights` isn't taken, I would suggest renaming your function to that.

lucasmoura:
Good suggestions, I will work on them. Thanks :)

MrDupin:
There are some conflicts now. May you resolve them?

lucasmoura:
@MrDupin Done

MrDupin:
Thanks.

PR #443, opened 2017-03-29
Notebook: Neural network

lucasmoura:
This MR add the Neural Network or MultiLayer Perceptron entry to learning.ipynb. It also explains a little bit about Backpropagation, but the section will still definitely need some improvements.


MrDupin:
Great work. There are some technical issues though. Here are a few pointers after a quick skim over your work (you can use Ctrl+F to find them):

* You seem to have duplicated Perceptron's code. Right above the example.

* The first sentence ("Although... linear classifier.") ran for too long. I suggest you break it up at the last comma.

* "A perceptron can be used" - Since you are capitalizing "Perceptron" in all other occasions, you should capitalize this instance too for consistency.

* "which is nonlinear function" - There should be an 'a' before 'nonlinear'.

* "Perceptron(MLP)" - Add a space before the opening parenthesis.

* "Different from the Perceptrom" - You accidentally typed "Perceptrom" with an 'm'.

* "this network also have a input" - It's "has" instead of "have".

* "An usual execution of a MLP" - It's 'A usual' and 'an MLP'. The word 'usual' begins with a 'you' sound and 'MLP' with an 'em' sound.

* "is feed to the next" - It's 'fed' instead of 'feed'.

* "(We currently only support" - The word 'We' should not be capitalized, even though it begins a sentence, since it is in parenthesis.

* "which will will receive" - You have two 'will' in a row.

* "As the PerceptronLearner" - It's "Like the PerceptronLearner".

* "NeuralNetLearner is is a binary" - Two "is" in a row.

lucasmoura:
Such silly mistakes. Thanks again for the input :)

MrDupin:
No problem.

Also, can you update the contents with the sections you added?

norvig:
This is very nice, but unfortunately we have a conflict now over the file, and it is not easy to resolve in .ipynb files.

MrDupin:
Hello @lucasmoura. I would hate for all this work to go to waste. If you don't have the time to resolve the conflicts, may I work on this?

lucasmoura:
@MrDupin Unfortunetely, I will not have time to work on this notebook right now. But if you want to update or rewrite what I have done, that would be awesome :)

MrDupin:
Great, thanks a lot!

norvig:
Closing this because #532 resolves the conflicts.

PR #445, opened 2017-03-30
Implementation: Transition Model for MDP

MrDupin:
Added a Transition Model to the `MDP` class.

The model is provided by the caller and is a dictionary of dictionaries in the form: `transitions[state][action] = (p, s)`. Now the function `T(self, action, state)` can return `self.transitions[state][action]`. Each `T(s, a)` call returns a tuple `(p, s')`, as per the guidelines in the comments.

Also, added tests for a basic Transition Model.


MrDupin:
*closing and re-opening to pass checks*

PR #455, opened 2017-03-31
Changed max to argmax

Chipe1:
Changed max() to argmax() to make code easier to read.

### DIFF
--- a/text.py
+++ b/text.py
@@ -318,7 +318,7 @@ class ShiftDecoder:
     def decode(self, ciphertext):
         """Return the shift decoding of text with the best score."""
 
-        return max(all_shifts(ciphertext), key=lambda shift: self.score(shift))
+        return argmax(all_shifts(ciphertext), key=lambda shift: self.score(shift))
 
 
 def all_shifts(text):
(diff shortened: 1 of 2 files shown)


text.py
```diff
@@ -380,8 +378,9 @@ def __init__(self, initial=None, goal=None, decoder=None):
 
     def actions(self, state):
         # Find the best
-        p, plainchar = max([(self.decoder.P1[c], c)
-                            for c in alphabet if c not in state])
+                            
+        search_list = [c for c in alphabet if c not in state]
+        plainchar = argmax(search_list, key=lambda c: self.decoder.P1[c])
```
norvig: This is fine, but I wonder which is clearer:
     lambda c: self.decoder.P1[c]
or
     self.decoder.P1.get

norvig:
Sorry, I was late at getting to this, and now there is a conflict; can you fix?

text.py
Chipe1: Did you mean self.decoder.P1.__getitem__
I got an error when I tried self.decoder.P1.get `AttributeError: 'UnigramTextModel' object has no attribute 'get'`

Chipe1:
Updated PR in #481

PR #457, opened 2017-03-31
added policy_iteration to mdp NOTEBOOK

sofmonk:
added policy_iteration to mdp notebook

### DIFF
--- a/mdp.ipynb
+++ b/mdp.ipynb
@@ -17,7 +17,7 @@
    },
    "outputs": [],
    "source": [
-    "from mdp import MDP, GridMDP, sequential_decision_environment, value_iteration"
+    "from mdp import *"
    ]
   },
   {
@@ -230,7 +230,7 @@
     {
      "data": {
       "text/plain": [
-       "<mdp.GridMDP at 0x7f256c52c390>"
+       "<mdp.GridMDP at 0x7f33a45a3ac8>"
       ]
      },
      "execution_count": 7,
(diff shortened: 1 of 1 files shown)


Chipe1:
I'm not sure what this adds to the notebook. I could see some '\n' removed, 'Widget Javascript not detected' errors and code from the mdp.py
Not sure why there are JS errors. We must use %psource to present code from the repo because any change in the code would need to be explicitly copied to the notebook otherwise.
Does anyoune know of a better method to view changes in a notebook. Looking at the diff on github is quite confusing.

sofmonk:
@Chipe1 You may click on the **view** button on the _files changed page_, though it doesn't highlight changes, but you can compare with previous notebook to see what was added.

mdp.ipynb
```diff
@@ -467,6 +483,102 @@
...
+    "The `policy_iteration(mdp)` function follows closely from fig. 17.7 pseudocode.\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def policy_iteration(mdp):\n",
```
Chipe1: Use %psource instead of pasting the code from mdp.py

mdp.ipynb
```diff
@@ -467,6 +483,102 @@
...
+    "Given a policy `pi` , the `policy_evaluation` function calculates the utility of each state if `pi` were to be executed. The i<sup>th</sup> iteration specifies the action to be taken in state `s`. We can perform any no. of iterations (`k=` 50 here) to give a reasonably good approximation of utilities."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 16,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def policy_evaluation(pi, U, mdp, k=50):\n",
```
Chipe1: Use %psource instead of pasting the code from mdp.py

mdp.ipynb
```diff
@@ -435,14 +423,42 @@
    },
    "outputs": [
     {
+     "name": "stderr",
```
Chipe1: Can these errors be avoided?
I get the same error but only once. Maybe removing it through text editor could work as a temporary fix

mdp.ipynb
norvig: I agree -- we need to use %psource so that we don't have the same code repeated in two places.

norvig: I'm closing this for now.

PR #293, opened 2017-03-01
Added some tests in test_learning.py

theSparta:
Added tests for the `error` functions in `learning.py` as required by #286 .
@reachtarunhere, What kind of other tests can be added?

PR #458, opened 2017-03-31
Add new tests to learning.py

lucasmoura:
Add tests for the following functions in learning.py

* rms_error
* ms_error
* mean_error
* manhattan_distance
* mean_boolean_error
* hamming_distance

### DIFF
--- a/tests/test_learning.py
+++ b/tests/test_learning.py
@@ -1,7 +1,114 @@
+import pytest
+
 from learning import parse_csv, weighted_mode, weighted_replicate, DataSet, \
                      PluralityLearner, NaiveBayesLearner, NearestNeighborLearner, \
-                     NeuralNetLearner, PerceptronLearner, DecisionTreeLearner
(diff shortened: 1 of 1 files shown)


Chipe1:
Most of the tests were already added in #293. Line 60 throws "AttributeError: 'module' object has no attribute 'approx'"

tests/test_learning.py
```diff
@@ -1,7 +1,114 @@
...
+
+    predictions = [1, 1, 1]
+    targets = [2, 2, 2]
+
+    assert rms_error(predictions, targets) == 1
+
+    predictions = [1, 1, 2, 1]
+    targets = [1, 2, 1, 1]
+
+    assert rms_error(predictions, targets) == pytest.approx(sqrt(0.5))
```
sofmonk: `pytest` will have to be removed.

lucasmoura:
@Chipe1 Thanks for pointing out about the other MR request. I have also realized that I am using a newer version of py.test than the one in the build. But since there is another MR covering the same tests as these ones, I will close it for now.

PR #460, opened 2017-03-31
Distance Functions (Euclidean+ Notebook)

MrDupin:
I added the distance functions to the Notebook + I implemented the Euclidean Distance.

Also, I renamed the arguments in the functions. Previously they were "prediction" and "targets", but I believe the more general "X" and "Y" are better. Especially considering that in some cases the two items compared are technically not a prediction and a target. Like in the *k-Nearest Neighbors* algorithm, where we are simply calculating the distance between two items.

Finally, added tests for the `euclidean_distance` function.

### DIFF
--- a/learning.ipynb
+++ b/learning.ipynb
@@ -14,7 +14,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 6,
    "metadata": {
     "collapsed": true,
     "deletable": true,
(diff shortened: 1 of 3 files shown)


Chipe1:
Adding separate test cases won't cause merge conflicts. You can go ahead and add tests for euclidean distance.

MrDupin:
I believe the conflict won't be in the test itself, but in the imports of the file. I will write the tests either way, and I will resolve any conflict afterwards since it's not going to be too much work.

Thanks.

PR #456, opened 2017-03-31
Implemented PermutationDecoder

Chipe1:
Added a hashable dictionary to allow dicts to act as states of a graph

PR #463, opened 2017-04-01
Added ShiftDecoder to notebook

Chipe1:
I will add permutation decoder after #456 is merged.

### DIFF
--- a/text.ipynb
+++ b/text.ipynb
@@ -15,9 +15,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 45,
+   "execution_count": 1,
    "metadata": {
-    "collapsed": true
+    "collapsed": true,
+    "deletable": true,
+    "editable": true
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


MrDupin:
Nice addition.

I will note some issues I found after a quick read:

I believe it is "Caesar cipher" - not "ceasar cipher". Aside from the misspelling, I believe it should be capitalized as it is named after a person. Also, in most of the writing (at least online) "Caesar" is capitalized.

I think there is a singular/plural issue in your text, but I'm not sure:

"These are a very basic form of cipher" - I'm not knowledgeable of the use of 'cipher' in scientific literature, but in normal text it is singular. So it should be 'ciphers', since you are talking about many of them.

A couple of other minor issues:

* "message text(also known as plaintext)" - There should be a space before the first parenthesis.

* "shifter by 3 yields" - It should be 'shifted'.

Chipe1:
@MrDupin Thanks for the review. I'm still not sure about the "These are a very basic form of cipher". I've changed the line to "These are primitive forms of cipher". Do suggest any improvement.

MrDupin:
The new line is better, thanks.

PR #465, opened 2017-04-01
added test to test_planning.py

sofmonk:
added the other possible solution to test for `air_cargo_problem`.

### DIFF
--- a/tests/test_planning.py
+++ b/tests/test_planning.py
@@ -18,17 +18,33 @@ def test_action():
     assert not a.check_precond(test_kb, args)
 
 
-def test_air_cargo():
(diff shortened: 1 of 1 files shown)


tests/test_planning.py
```diff
@@ -18,17 +18,33 @@ def test_action():
...
+                expr("Unload (C2, P2, SFO)")]  
 
-    for action in solution:
+    for action in solution_1:
+        p.act(action)
+
+    assert p.goal_test()
+
+
+def test_air_cargo_2():
```
Chipe1: It would be better to test both solutions in a single test_air_cargo()

PR #466, opened 2017-04-01
Added tests to test_utils.py

sofmonk:
Added tests for` count()` and `mode()`.


MrDupin:
I don't feel the tests for `count` are representative of the function. In the cases you provided, `count` works just as `len`, returning the length of the input. In my opinion a better test would be this:

`count([True, False, True, True, False])`

The above has a length of 5, but `count` will return 3 since that's the number of elements which are `True` in the input. You can expand the above example to include conditions. For example:

`count([5 > 1, len("abc") == 3, 3+1 == 5])`

This better showcases the function.

PR #478, opened 2017-04-03
implemented Genetic Algorithm in search.py and made modifications acc…

Agent-Pandit:
…ording to flake8. Demonstrated the working of GA in search.ipynb.


Chipe1:
Not sure why we have __name__ == "__main__" in the notebook cell.
It would be better if we removed the large print statements. They are too big. Maybe print every 100th line.
There is an images folder which stores all images used in the notebook. Hotlinking the file is not a good idea as the source may fail and we cannot view the notebook offline.

Agent-Pandit:
@Chipe1 I agree that name == "main" serves no purpose and can be removed.

Regarding print statements, I believe that it gives an opportunity to analyse the evolution of string, moreover it is also suggested to modify the code (in the notebook). So after one has understood the basic concepts of GA using the notebook, one can modify the print statement to satisfy the particular need. Our purpose is not to build the most optimised search machine but to demonstrate the working of the algorithms. The print statements, therefore is useful. This is my personal point of view If the community feels that we should do otherwise we will do otherwise.

About images… I agree with you…
ModuleNotFoundError: No module named 'grid' - not sure about this.

Chipe1:
@Agent-Pandit My main concern is the output size. It makes the file unnecessarily large - 1.8MB of text, which can be generated by a simple call to a function . It would be better if mentioned how to generate the sequence with a warning that the output would be very large, maybe comment the print statements and ask the viewers to uncomment it and run the code to see the evolution of string. Storing such a large output doesn't look like a good idea to me. Is there any other example in which the sequence is small, say 100 lines or less. Using small examples might be better and easier to digest.

norvig:
I agree with @Chipe1 -- let's print every N lines, and have N default to, say, 100, but mention how to change it.

PR #483, opened 2017-04-07
Tests: Extend Learning Tests

MrDupin:
Added more thorough testing in the `learning` module for the kNN and DecisionTreeLearner algorithms. The tests now include items from all classes.


tests/test_learning.py
```diff
@@ -41,13 +41,17 @@ def test_k_nearest_neighbors():
 
     kNN = NearestNeighborLearner(iris,k=3)
     assert kNN([5,3,1,0.1]) == "setosa"
+    assert kNN([6,5,3,1.5]) == "versicolor"
```
norvig: Not a big deal, but there should be a space after commas.

PR #484, opened 2017-04-07
Fix learning tests

lucasmoura:
Two tests on test_learning.py are failing now. That's because they are testing code that still have not been merged. Both tests address the following MR: #435.

This MR simply remove these tests.

However, I do believe that both of these tests should be added to #435.

### DIFF
--- a/tests/test_learning.py
+++ b/tests/test_learning.py
@@ -35,20 +35,6 @@ def test_weighted_replicate():
     assert weighted_replicate('ABC', [1, 2, 1], 4) == ['A', 'B', 'B', 'C']
 
 
-def test_means_and_deviation():
-    iris = DataSet(name="iris")
-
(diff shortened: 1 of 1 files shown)


MrDupin:
The tests you mentioned *are* added in #435.

I unfortunately added the tests in another PR, which I didn't think would get merged before #435. My bad.

lucasmoura:
No problem, it happens :)

MrDupin:
I just want to note that this PR should be merged **only** if #435 is declined. Otherwise there is no point, and might possibly cause conflicts with other PRs dealing with `test_learning`.

norvig:
Thanks for this -- my mistake in taking a few days off and getting behind in merges, and taking them in the wrong order. Glad to have this fix!

PR #486, opened 2017-04-07
Implementation: Multi-Class Backpropagation

MrDupin:
The `BackPropagationLearner` can now learn for more than two classes. Previously, it was a binary learner, so `NeuralNetworkLearner` and `PerceptronLearner` were binary classifiers. I extended the functionality to classify into multiple classes.

* Added a new function, `init_examples`, to initialize the inputs and outputs for the learner. The outputs need to be in an one-hot representation, where the 1 represents the class the item is classified in.

* Added tests for the algorithms (unfortunately, the learners can sometimes be wrong, so the tests are a bit general).

* Added a new function, `find_max_node`, which takes as input the output nodes and returns the index of the one with max value.

* Moved the helper functions below the three main algorithms (Neural Network, BackPropagation and Perceptron).

* Implemented multi-class functionality.

**Note:** The `learning` Notebook needs to be updated too, but I would prefer to do it after this is reviewed so we are certain this is the implementation we want to go with.

### DIFF
--- a/tests/test_learning.py
+++ b/tests/test_learning.py
@@ -52,23 +52,33 @@ def test_decision_tree_learner():
 
 def test_neural_network_learner():
     iris = DataSet(name="iris")
-    iris.remove_examples("virginica")
-    
-    classes = ["setosa","versicolor","virginica"]
-    iris.classes_to_numbers()
 
-    nNL = NeuralNetLearner(iris)
-    # NeuralNetLearner might be wrong. Just check if prediction is in range.
-    assert nNL([5,3,1,0.1]) in range(len(classes))
+    classes = ["setosa","versicolor","virginica"]
+    iris.classes_to_numbers(classes)
+
+    nNL = NeuralNetLearner(iris, [5], 0.15, 75)
+    pred1 = nNL([5,3,1,0.1])
+    pred2 = nNL([6,3,3,1.5])
+    pred3 = nNL([7.5,4,6,2])
+
+    # NeuralNetLearner might be wrong. If it is, check if prediction is in range.
+    assert pred1 == 0 or pred1 in range(len(classes))
+    assert pred2 == 1 or pred2 in range(len(classes))
+    assert pred3 == 2 or pred3 in range(len(classes))
 
 
 def test_perceptron():
     iris = DataSet(name="iris")
-    iris.remove_examples("virginica")
-    
-    classes = ["setosa","versicolor","virginica"]
     iris.classes_to_numbers()
 
+    classes_number = len(iris.values[iris.target])
+
     perceptron = PerceptronLearner(iris)
-    # PerceptronLearner might be wrong. Just check if prediction is in range.
-    assert perceptron([5,3,1,0.1]) in range(len(classes))
+    pred1 = perceptron([5,3,1,0.1])
+    pred2 = perceptron([6,3,4,1])
+    pred3 = perceptron([7.5,4,6,2])
+
+    # PerceptronLearner might be wrong. If it is, check if prediction is in range.
+    assert pred1 == 0 or pred1 in range(classes_number)
+    assert pred2 == 1 or pred2 in range(classes_number)
+    assert pred3 == 2 or pred3 in range(classes_number)
--- a/learning.py
+++ b/learning.py
@@ -465,7 +465,7 @@ def NeuralNetLearner(dataset, hidden_layer_sizes=[3],
     """
 
     i_units = len(dataset.inputs)
-    o_units = 1  # As of now, dataset.target gives only one index.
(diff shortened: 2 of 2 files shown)


lucasmoura:
Cool work, just one suggestion, the function **find_max_node** could be rewritten to:

```python
def find_max_node(nodes):
    return nodes.index(max(nodes, key=lambda node: node.value))
```

MrDupin:
I was thinking of writing it like this, which is more pythonic, but I don't feel it's that needed. First, we are writing a lot of keywords (`index`, then `max`, then the `lambda`) all in one line which isn't that easy to the eye. Second, this is a simple helper function that everyone knows what it does by the name alone. The way I have currently written it is the way that everyone has been seeing it for a long, long time. I just imagine that if someone was skimming through the function, he wouldn't mind it the way it is written now, but if it was written in the pythonic way you mentioned, it might catch an eye or two.

Anyway, this was my reasoning behind this decision. Long story short, even though there is a more pythonic way to write it, the function itself is so commonly seen the way I wrote it that I think it balances out.

I don't have strong feelings on this though and I would change it if more people feel the same way. Thanks for the suggestion.

Chipe1:
I too feel we need to change `find_max_node` function. The way it uses index as -1 and initial maximum as -10000 feels 'hackish'. Also use `argmax` instead of `max`. It is the standard notation used in the book.

MrDupin:
Since you agree with @lucasmoura too, I will make the changes.

This is what I will use:

```python
def find_max_node(nodes):
    return nodes.index(argmax(nodes, key=lambda node: node.value))
```

Unfortunately my PC burned down so I can't test this. Can someone confirm the above returns the index of the max element? Thanks.

lucasmoura:
@MrDupin I have made the following test and it works fine:

```python
from collections import namedtuple
from utils import argmax
import pytest


def test_node():
    Node = namedtuple('Node', ['value'])

    nodes = [Node(value) for value in range(10)]

    assert 9 == nodes.index(argmax(nodes, key=lambda node: node.value))
```

MrDupin:
I just thought of something else that maybe should be discussed.

I changed `max_score` to 0 (at least for now). If the values in the nodes don't exceed 0, something went terribly wrong. Since the `BackPropLearner` converges in values in the range [0,1], if the maximum value is negative, there is an issue. If `max_score` is initialized at 0, we will catch that issue but in the pythonic way we won't. In that case, a seemingly random index will be returned and the user will be none the wiser.

So, a new idea I want to throw on the table is this: We keep the traditional find max method that I wrote, but if the index at the end is -1, we raise an exception.

Another idea is to go the pythonic route, but in the end check if `max_score` is negative. If yes, we raise an exception. I don't feel though this is as pythonic and beats the purpose of that route.

I would like another opinion on this. Sorry for taking so long to decide, this is a pretty big PR and I wouldn't want to mess it up.

---

PS: The title of the second to last commit is wrong, I rushed it. 0 is better than -1.

Chipe1:
@MrDupin Let's change it to the pythonic implementation and add the >0 check in test files. If there is something wrong in the implementation, it must ideally be caught in tests and fixed. I don't like the idea of performing runtime checks. They must be used only for debugging. Adding these checks makes the code hard to read and defeats the purpose of tests.

tests/test_learning.py
```diff
@@ -52,23 +52,33 @@ def test_decision_tree_learner():
...
+    nNL = NeuralNetLearner(iris, [5], 0.15, 75)
+    pred1 = nNL([5,3,1,0.1])
+    pred2 = nNL([6,3,3,1.5])
+    pred3 = nNL([7.5,4,6,2])
 
-    nNL = NeuralNetLearner(iris)
-    # NeuralNetLearner might be wrong. Just check if prediction is in range.
-    assert nNL([5,3,1,0.1]) in range(len(classes))
+    # NeuralNetLearner might be wrong. If it is, check if prediction is in range.
+    assert pred1 == 0 or pred1 in range(len(classes))
```
norvig: This is ok for now. But in the long run, two things:
(1) There should be a way to test any learner, not just NeuralNetLearner
(2) To deal with the possibility of nondeterministic learners getting it wrong, there shoud be experiments, with overall results graded (for example, mst get at least 2 out of 3 answers right).

MrDupin: Thanks for the feedback, I appreciate it.

1) I'm not sure I fully understand this, is it possible to elaborate? I could write a function which, receiving a learner as input, will run the input learner and return the result, if that's what you are asking.

2) That is a great idea! I will get to work.

MrDupin: I have implemented 2) on #496. I hope this is what you were asking for.

PR #487, opened 2017-04-07
Temporarily remove flake8 tests for build to pass

Chipe1:
I feel we should add flake8 to travis only after the repo in its current state has no flake8 errors. There are many PRs to be merged which may cause the build to fail(according to travis). Travis will still run flake8 checks but that wont affect build status.
 I suggest we first wait till the important PRs are merged, fix any flake8 errors and only then make flake8 compulsory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ install:
 script:
   - py.test
   - python -m doctest -v *.py
-  - flake8 .
 
 after_success:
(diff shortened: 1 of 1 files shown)


norvig:
Good idea! And in general, I think of style guides as "guidelines", not "rules", and sometimes it is violating a guideline is better than contorting code to obey it. So when we bring flake8 back, it is fine to disable some of the warnings, and clean them up manually rather than automatically.

PR #488, opened 2017-04-09
Implemented GA in notebook search.ipynb

Agent-Pandit:
Removed the image links and included the images used in the image folder. Reduced the file size from 2.4 MB to 183 KB. For every 21 print statements that we previously had, now we have 2, only printing the fittest individual in each generation. The instructions has been made to get the detailed output.


Chipe1:
@Agent-Pandit You might have to add .idea in your gitignore or not add it in you commit. There are a few xml files which I think are produced by your IDE. It also has new line as '\r\n' instead of '\n' as mentioned in #417

Agent-Pandit:
@Chipe1 thanks for pointing it out. Do I need to make another PR or this will work?

Chipe1:
I'm not sure how to uncommit changes. Maybe you can delete the files in a new commit and also fix the merge conflict.

PR #496, opened 2017-04-13
Learning: Grade Learner

MrDupin:
Implemented a function to grade a learner on how many tests it passes. Given a learner and a list of tests, it counts how many tests the learner got right.

In `test_learning`, replaced the tests of the non-deterministic algorithms with the new function. If the learner guesses 2/3 outputs correctly, it passes. (I find it extremely unlikely, near impossible, for `test_learning` to fail.)

Finally, I spaced out elements in list to comply with PEP8.

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -908,6 +908,16 @@ def learningcurve(learner, dataset, trials=10, sizes=None):
     return [(size, mean([score(learner, size) for t in range(trials)]))
             for size in sizes]
 
+
+def grade_learner(predict, tests):
+    """Grades the given learner based on how many tests it passes.
+    tests is a list with each element in the form: (values, output)."""
+    correct = 0
+    for t in tests:
+        if predict(t[0]) == t[1]:
+            correct += 1
+    return correct
+
 # ______________________________________________________________________________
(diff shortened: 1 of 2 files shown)


lucasmoura:
I believe that the **test** function in learning.py could be used for that. The main difference is that it returns the amount of incorrect predictions, instead of the number of correct ones.

learning.py
```diff
@@ -908,6 +908,16 @@ def score(learner, size):
     return [(size, mean([score(learner, size) for t in range(trials)]))
             for size in sizes]
 
+
+def grade_learner(predict, tests):
```
norvig: I think this would be better if it returned the ratio (i..e. percent) of correct tests, rather than the total.
Then we can call it with `grade_learner(predicter, tests) > 2/3`, regardless of length of tests.

Also, `for t in tests` should be `for (X, y) in tests`
Or the whole thing could be `return mean(int(predict(X) == y) for X, y in tests)`

MrDupin:
Thanks for the feedback, much appreciated.

I made the suggested changes in #499.

PR #499, opened 2017-04-14
Learning: Grading Learners

MrDupin:
*The "sequel" to #496.*

After the feedback from @lucasmoura and @norvig, I changed the previously introduced function `grade_learner`. It now returns the ratio of correct answers.

I chose not to use the `test` function since it is tailored to the evaluation of a learner with the fold validation method (bulk error-finding in the folds). I believe it would be best if we also had a simple function to indicate loosely and quickly how good a learner is.

So I wrote the simpler function `grade_learner` to check a learner against a set of examples (in the form of tuples).

Also, I renamed `test` to `err_ratio`. Unfortunately **pytest** mistakes the name for a function it needs to test and throws an error. `err_ratio` does not have that issue.

Finally, I used the new function to test the Neural Network and the Perceptron.

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -806,8 +806,9 @@ def flatten(seqs): return sum(seqs, [])
 # Functions for testing learners on examples
 
 
-def test(predict, dataset, examples=None, verbose=0):
+def err_ratio(predict, dataset, examples=None, verbose=0):
     """Return the proportion of the examples that are NOT correctly predicted."""
+    """verbose - 0: No output; 1: Output wrong; 2 (or greater): Output correct"""
     if examples is None:
         examples = dataset.examples
     if len(examples) == 0:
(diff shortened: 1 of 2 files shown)


lucasmoura:
@MrDupin I think I have not explained my idea properly, and I am sorry for that. I was thinking about using the **test** function after training the model and using to evaluate the learner. Currently, you are using modified test cases to evaluate the learner right ? I think that what you can do is to evaluate the whole trainining process, and not just a few examples. And for that, use the **test** function.

But again, maybe this is not the best approach :)

MrDupin:
Your suggestion is definitely one way to go about it and I don't think there is right and wrong in this case. In my opinion a simple function to run a learner through some tests is useful to have around. The `test` function is mostly for the bulk error-checking used in operations like the fold validation and for a more robust evaluation of a learner (with the option to print out as it goes along).

I think I will add the `test` in `test_learning` though, since we need to be checking that one too. I will have the learner evaluated with the function on the whole iris dataset.

Having both is the best way to go, in my opinion.

PR #500, opened 2017-04-14
Fix build

lucasmoura:
The build is currently broken, that's because the fuzzywuzzy lib was not being installed during the build.

Also, there was a error message with a local path in the search notebook, which I think is a bad idea. Therefore, I removed it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
+fuzzywuzzy==0.15
 networkx==1.11
 jupyter
--- a/search.ipynb
+++ b/search.ipynb
@@ -22,16 +22,7 @@
     "editable": true,
     "scrolled": true
    },
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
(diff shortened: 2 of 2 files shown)


MrDupin:
I agree with @Chipe1 on #497 that it would be better if we did not use the fuzzywuzzy library just for this one function, so I don't think we should be adding it to the requirements.

lucasmoura:
@MrDupin Yes, I agree adding the new lib is not a good idea. However, until we fix that properly, I don't think the build should stay broken. That's the only reason for this MR.

MrDupin:
Yes, definitely.

Chipe1:
@lucasmoura #501 would fix the dependency issue. As for the error with the local path, there are 4 more in search.ipynb
Would be nice if you removed those errors too.

MrDupin:
I fixed the other notebook errors in another PR, #503.

lucasmoura:
Seems this MR is not necessary anymore.

PR #501, opened 2017-04-14
Implementation: Genetic Algorithm (Fixing Build)

MrDupin:
After #480 got merged, the implementation veered away from the pseudocode in the book, plus it was a bit restricting in the problems we could solve.

I reverted the implementation to (barring a few changes to adhere to pseudocode more closely) the previous one. Also, I implemented functions `reproduce` and `mutate` according to the book + `init_population` to make initializing a random population easier.

I removed the `GA_State` class from the original implementation, since we can do without it while also making the code closer to the book.

Finally, added tests for a small graph coloring problem, including problem basis and fitness function. I have tested it 100 times to make sure the algorithm gives the correct result, but unfortunately due to the nature of the algorithm we can't be sure it will always produce the correct solution (it is heavily based on luck). Still, I believe it is very unlikely that the test will fail, and we can always increase population and generations if need be.

UPDATE: Added test for the 8-Queens Problem. Unfortunately it takes a long time to find the best solution (~5 minutes) so I put a fitness threshold of 25 (out of 28), which is very fast. The implementation fares well for 26 too, but it takes a bit more so I stuck with 25.

UPDATE 2: Added a section in the Notebook. Since both PRs are quite large, I decided to split them. The other PR is #503.

---
NOTE: Currently I have only tested the code with the general case of string populations. In the future we could try and make the algorithm also work with the Problems we introduce in the module (like the class `NQueensProblem`), using the `genetic_search` function.

### DIFF
--- a/search.py
+++ b/search.py
@@ -605,17 +605,17 @@ def genetic_algorithm(population, fitness_fn, gene_pool=['0', '1'], f_thres=None
     return argmax(population, key=fitness_fn)
 
 
-def init_population(pop_number, gene_pool, length):
+def init_population(pop_number, gene_pool, state_length):
     """Initializes population for genetic algorithm
-    pop_number:  Number of individuals in population
-    gene_pool :  List of possible values for individuals
-                 (char only)
-    length    :  The length of each individual"""
+    pop_number  :  Number of individuals in population
+    gene_pool   :  List of possible values for individuals
+                   (char only)
+    state_length:  The length of each individual"""
     g = len(gene_pool)
     population = []
     for i in range(pop_number):
         new_individual = ''.join([gene_pool[random.randrange(0, g)]
-                                  for j in range(length)])
+                                  for j in range(state_length)])
         population.append(new_individual)
 
     return population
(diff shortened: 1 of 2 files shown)


search.py
```diff
@@ -572,94 +569,74 @@ def LRTA_cost(self, s, a, s1, H):
...
+        new_individual = ''.join([gene_pool[random.randrange(0, g)]
+                                  for j in range(state_length)])
+        population.append(new_individual)
+
+    return population
+
+
+def reproduce(x, y):
+    n = len(x)
+    c = random.randrange(0, n)
```
Chipe1: Using randrange(0, n) makes the reproduction unfair since we can return unaltered `y` (when c is 0) but not `x`. The issue is very minor but makes the output biased towards `y` nonetheless.

MrDupin: Nice catch, thanks. Randrange now picks in (1, n).

norvig:
This is good, @MrDupin, thanks. In general, we would like to keep the code close to the pseudocode in the book. In cases where we would like to add additional functionality, it is best to add a new function, rather than complicate functions that mimic the pseudocode.

PR #506, opened 2017-04-17
Planning notebook

kaivalyar:
The planning notebook was empty, so I went ahead and added some examples using a subset of the familiar Romania graph from the search notebook.

### DIFF
--- a/planning.ipynb
+++ b/planning.ipynb
@@ -15,7 +15,7 @@
    "source": [
     "This notebook describes the [planning.py](https://github.com/aimacode/aima-python/blob/master/planning.py) module, which covers Chapters 10 (Classical Planning) and  11 (Planning and Acting in the Real World) of *[Artificial Intelligence: A Modern Approach](http://aima.cs.berkeley.edu)*. See the [intro notebook](https://github.com/aimacode/aima-python/blob/master/intro.ipynb) for instructions.\n",
     "\n",
-    "We'll start by looking at `PDDL` and `Action` data types for defining problems and actions. Then, we will see how to use them by trying to plan a trip across the familiar map of Romania, from [search.ipynb](https://github.com/aimacode/aima-python/blob/master/search.ipynb). Finally, we will look at the implementation of the GraphPlan algorithm.\n",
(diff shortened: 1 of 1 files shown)


MrDupin:
Good work. You described the concepts nicely.

A few pointers you can improve on:

* "The first step is to load the code ... Lets first think about modelling an Action" - You write 'first' two times in a row and it doesn't look nice. I would suggest rephrasing the second sentence.

* When you have pointers, I suggest you capitalize them.

* A lot of times you write "lets". It actually is "let's", with an apostrophe before 's'.

* Usually small numbers in sentences are spelled out instead of writing them in numerals. For example: "We need at least 3 things" would be "We need at least three things". Note that "Chapter 10" is correct. This is a small thing and is a matter of preference, but in other Notebooks we mostly spell them out. Personally I would change it for consistency, but it's up to you.

* "This is to workaround the fact" - You should space "workaround" to "work around". You use it as a verb in the sentence, and "workaround" is not a verb.

* "Similarly, we track the effects in terms of the statements that become true, should an action be executed, and those that become false, once the action is executed." - This needs a bit of work. The 'should an action be executed' and 'once the action is executed' parts are a bit confusing. From my understanding they mean the same thing, but the first might also mean "in case an action is executed" while the second one might mean "after an action is executed". I believe they mean the same thing, but you changed the wording to avoid repetition. To avoid confusion and repetition, you can write something like this: "Similarly, we track the effects in terms of the statements that become true and those that become false once the action is executed."

* "Next, actions contains a list of `Action` objects that may" - I'm not sure if "actions" is a variable, but if it isn't it should be followed by "contain" instead of "contains". If it is a variable, I would style it differently to set it apart.

* "But, as clear from this list of airports" - I think there's a verb missing here. The sentence is not structured properly. Maybe an 'is' after 'as'?

* You use the words "now" and "let's" a lot. I would suggest you rephrase some of these to make the text flow more smoothly.

* You end the tutorial by defining a variable `prob`, but you don't do anything with it. This can of course be added later, but I would like to see something more done with it.

sofmonk:
PR #427. I created a PR for planning.ipynb too. 
We are going to have some merge conflicts.

kaivalyar:
@sofmonk Sorry I didn't anticipate that. Are you sure the code in your notebook will still work? There have been substantial changes in planning.py itself since #427 - eg. PDLL is now PDDL. Also, I haven't yet done any of the air-cargo problem or spare-tire problem in the notebook as you had, in #427. Maybe we could merge those in after you have refactored your notebook to reflect the updated planning.py module?

kaivalyar:
Thanks for the exhaustive suggestions, MrDupin. I have tried to incorporate most of them. The notebook reads much better than before now.

norvig:
Nice work @kaivalyar, and helpful comments @mrdupin.

PR #521, opened 2017-05-28
CSP Notebook: Outputs

MrDupin:
I show previously hidden outputs. Also, removed matplotlib warnings. Javascript widget error warnings remain.

EDIT:

Note that in one of the outputs, all that is printed is pointers to some objects. I'm not sure what the purpose of this is. I left them in, but personally I feel that that output should be hidden, or maybe have the cell removed completely.

### DIFF
--- a/csp.ipynb
+++ b/csp.ipynb
@@ -2,11 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {
-    "collapsed": false,
-    "deletable": true,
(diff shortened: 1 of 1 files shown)


norvig:
I agree that "In[3] australia, usa, france: adds no value. We should either delete that cell, or add a useful __str__/__repr__ method to CSP.

It is depressing that we can't get the javascript to load in github. I'm not sure what a better solution is.

PR #522, opened 2017-05-28
Search Notebook: Fixing Latex

MrDupin:
Latex in the Genetic Algorithm section was not compiling. After some probing I found that the issue was the `limits` code. I worked around it. The end result, in my opinion, is not very pretty, but it is understood and gets the job done.


norvig:
Good detective work!

Issue #502, opened 2017-04-16
Weird Issue with Weighted Sampler and Travis


MrDupin:
Hi guys. I have encountered an issue with the Travis checker that I for some reason can't replicate on my machine. I would like some help, if possible.

I recently rewrote the implementation of the genetic algorithm. The PR is #501. The code for the core algorithm was implemented (mostly, I did make some changes) by someone else before me. Here's the code (I blanked the arguments, since it's too long for Github comments):

```
def genetic_algorithm( ... ):
    """[Figure 4.8]"""
    for i in range(ngen):
        new_population = []
        for j in range(len(population)):
            fitnesses = map(fitness_fn, population)
            p1, p2 = weighted_sample_with_replacement(2, population, fitnesses)
            child = reproduce(p1, p2)
            if random.uniform(0, 1) < pmut:
                child = mutate(child, gene_pool)
            new_population.append(child)

        population = new_population

        if f_thres:
            fittest_individual = argmax(population, key=fitness_fn)
            if fitness_fn(fittest_individual) >= f_thres:
                return fittest_individual

    return argmax(population, key=fitness_fn)
```

I would like to bring your attention to the following lines:

`fitnesses = map(fitness_fn, population)`

and

`p1, p2 = weighted_sample_with_replacement(2, population, fitnesses)`

With these lines we pick randomly the most fit individuals. This line is called in each iteration. As it is right now there are no issues. Since though `population` doesn't change inside each generation, I thought it would be best to put the line outside the nested for loop, like this:

```
for i in range(ngen):
    fitnesses = map(fitness_fn, population)
    new_population = []
    for j in range(len(population)):
        p1, p2 = weighted_sample_with_replacement(2, population, fitnesses)
        ...
```

That way we save the time it takes to map fitnesses. When I ran the above in my machine, no errors are thrown. But when Travis checked, it actually threw the following error:

```
search.py:592: in genetic_algorithm

    p1, p2 = weighted_sample_with_replacement(2, population, fitnesses)

utils.py:202: in weighted_sample_with_replacement

    return [sample() for _ in range(n)]

utils.py:202: in <listcomp>

    return [sample() for _ in range(n)]

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   return lambda: seq[bisect.bisect(totals, random.uniform(0, totals[-1]))]

[...]

PR #523, opened 2017-05-29
Genetic Algorithm: String to List Individuals

MrDupin:
As per the discussion on #347, I changed the individuals from string to list. Updated tests accordingly.

I broke the Graph Coloring test into three mini-tests, one for each of the three main variable types (char/string, integer, boolean).

NOTE: I have updated the notebook too, but I am waiting till #522 gets merged to avoid dealing with conflicts.

### DIFF
--- a/search.py
+++ b/search.py
@@ -582,7 +582,7 @@ def genetic_search(problem, fitness_fn, ngen=1000, pmut=0.1, n=20):
     return genetic_algorithm(states[:n], problem.value, ngen, pmut)
 
 
-def genetic_algorithm(population, fitness_fn, gene_pool=['0', '1'], f_thres=None, ngen=1000, pmut=0.1):  # noqa
+def genetic_algorithm(population, fitness_fn, gene_pool=[0, 1], f_thres=None, ngen=1000, pmut=0.1):  # noqa
     """[Figure 4.8]"""
     for i in range(ngen):
         new_population = []
@@ -610,13 +610,11 @@ def init_population(pop_number, gene_pool, state_length):
     """Initializes population for genetic algorithm
     pop_number  :  Number of individuals in population
     gene_pool   :  List of possible values for individuals
-                   (char only)
     state_length:  The length of each individual"""
     g = len(gene_pool)
     population = []
     for i in range(pop_number):
-        new_individual = ''.join([gene_pool[random.randrange(0, g)]
-                                  for j in range(state_length)])
+        new_individual = [gene_pool[random.randrange(0, g)] for j in range(state_length)]
         population.append(new_individual)
 
     return population
@@ -635,7 +633,7 @@ def mutate(x, gene_pool):
     r = random.randrange(0, g)
 
     new_gene = gene_pool[r]
-    return x[:c] + new_gene + x[c+1:]
+    return x[:c] + [new_gene] + x[c+1:]
 
 # _____________________________________________________________________________
 # The remainder of this file implements examples for the search algorithms.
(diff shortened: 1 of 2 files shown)


kaivalyar:
@MrDupin Would it make sense to replace
`fitnesses = map(fitness_fn, population)`
`random_selection = weighted_sampler(population, fitnesses)`
and the 
`x = random_selection()`
`y = random_selection()`
lines with just a simpler
`x = random_selection(population, fitness-fn)`
`y = random_selection(population, fitness-fn)`
and define the `random_selection(population, fitness-fn)` elsewhere, differently?

I only suggest this because then the code will truly mirror the pseudocode, which is desirable in my opinion. It won't have the extra lines and different parameter lists like it does right now. I would be happy to make this change if you like, else you can just add it to this PR.

MrDupin:
@kaivalyar: This is more or less how I had it initially, but as per [Darius' suggestion](https://github.com/aimacode/aima-python/issues/502#issuecomment-294381574), I changed the implementation to the current version due to performance issues. It would take a very long time to complete due to the added computational cost. The difference in performance was quite astute and older machines could previously run only small and trivial-ish problems.

There are, as I see it, three main ways to go about it:

a) Keep it as is, choosing performance over two additional lines of code + the function parameters of `random_selection`.

b) Follow the pseudocode to the letter, writing some possibly ugly and harder to follow code on the side to relieve the computational pressure.

c) Keep it as is, but change the

```
fitnesses = map(fitness_fn, population)
random_selection = weighted_sampler(population, fitnesses)
```

part to something like:

```
random_selection = weighted_sampler(population, map(fitness_fn, population))
```

Which will reduce the additional lines to just 1. It is a bit uglier in my opinion though.

---

Whichever way we choose, we should definitely have `random_selection` calculated only once.

The first (current) and third options already accomplish that.

For the second to work, we have to define `random_selection` as a function that on its first call will compute `weighted_sampler` and return an item, but on any subsequent calls it won't recomputed the sampler, instead just returning an item. Note that there should be a way to "reset" this, so at every generation it actually takes into account the new fitnesses. That possibly means adding another line/altering the pseudocode, or making the function even more convoluted.

Personally, I believe we should leave it as is. The only difference from the pseudocode is the two additional lines to compute `random_selection` and the fact that the calls to `random_selection` inside don't pass parameters. I believe it is as readable as the pseudocode and the differences aren't that prominent.

kaivalyar:
Woah. This turned out to be a lot more complicated that I imagined. I think you're right, let's stick with the way things are currently. If we can find a solution that solves this issue, we can always update the code later.

In the meanwhile, let me explore if this can be solved differently somehow...

MrDupin:
No problem. If you need any help/feedback/discussion, reach out to me. I would love to help any way I can.

MrDupin:
There is another, more elegant solution, now that I think about it. Define a function that will calculate the fitnesses and return the `weighted_sampler`.

```
def selection_chance(fitness_fn, population)
    fitnesses = map(fitness_fn, population)
    return weighted_sampler(population, fitnesses)
```

And then inside the generations loop, replace

```
fitnesses = map(fitness_fn, population)
random_selection = weighted_sampler(population, fitnesses)
```

with `random_selection = selection_chance(fitness_fn, population)`. I believe this is overall a better solution. It makes the code closer to the pseudocode while also making it more readable.

I will make this change (assuming no other solution is found till then) once this PR gets merged.

norvig:
I think @MrDupin has a good solution for now. If/when we go to Python 3.6, there is a new [random.choices](https://docs.python.org/3/library/random.html) function which does a similar thing.

PR #525, opened 2017-05-31
Genetic Algorithm: Compliance to Pseudocode

MrDupin:
A small edit to make the code closer to the book. Now the code has only one additional line instead of two.

### DIFF
--- a/search.py
+++ b/search.py
@@ -586,8 +586,7 @@ def genetic_algorithm(population, fitness_fn, gene_pool=[0, 1], f_thres=None, ng
     """[Figure 4.8]"""
     for i in range(ngen):
         new_population = []
-        fitnesses = map(fitness_fn, population)
-        random_selection = weighted_sampler(population, fitnesses)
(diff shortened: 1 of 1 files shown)


norvig:
OK, this is an improvement, so I'll accept it. But I think maybe we should do this: In utils, define

def random_choices(population, weights=None, k=1)

in anticipation of Python 3.6 adoption of random.choices, and call this.

norvig:
Also, we should feel free to change the pseudocode (as well as the Python code) if we think it can be made clearer.

MrDupin:
@norvig: Thanks for the feedback. I was thinking of putting the function (`selection_chances`) in utils, but since it is basically a wrapper for `weighted_sampler` I was having second thoughts, especially since the function is tailor made for the Genetic Algorithm. I see the merit of putting it on utils (namely uncluttering search.py), but I think it's unlikely it will be useful to to other parts of the code (I could be wrong about this though).

I don't have strong feelings about this, so I leave it up to the rest of the community.

Issue #529, opened 2017-06-03
Games Notebook: Min-Max


MrDupin:
I have started work on the Min-Max Notebook section. Currently we show the reader how to play a game, comparing how the different types of algorithms we have (random, alpha-beta, min-max) operate. We do not go in-depth or explain the algorithms yet.

This is what I am working towards. I will explain how the Min-Max algorithm works, giving in-depth examples in addition to the current comparison.

norvig:
That's good. One tricky issue: should you assume that the reader will be using an interactive environment and can see turn-by-turn changes, or a static view of the notebook where everything has to be explicit.

MrDupin:
I will for now assume a static view (aka GitHub viewer). This will allow only for some simple examples, unfortunately. I will also leave some room in the section for something more interactive, for when we settle on a solution on the Notebook conundrum, so that the work can be built upon.

norvig:
I agree that is the best approach for now.

norvig:
It would be good to also look at grid.py/.ipynb, and decide if they need to be stand-alone, or if they can be folded back into utils.

I think they're in pretty good shape now. Only a few points:
* I don't like a[0] and a[1] -- I think that's not clear enough. I would prefer either of:
   - `x,y = a`
   - `X(a)`, where we have `def X(point): return point[0]; def Y(point): return point[1]`
* I would do `orientations = E, N, W, S = ((1, 0), (0, 1), (-1, 0), (0, -1))`
* I would do `turns = L, R = (+1, -1)`
* I'm not sure about naming conventions. PEP 8 says constants are all UPPER, but that is not always used. I'm not sure if the name should be `E` or `EAST` or `East`.

MrDupin:
@norvig: I believe the `grid` module should be placed in `utils`. Also, I don't think a notebook for `grid` is very useful. The code itself is self-explanatory, for anyone who wants to look further into this.

I prefer `x, y = a`, since it's a bit simpler.

MrDupin:
Added 1st half of the PR in #536.

norvig:
I agree that getting rid of gris.ipynb and moving grid functions to utils is the right thing.

MrDupin:
#539 adds the minimax algorithm to the notebook.

#540 moves grid to utils (also removes notebook + moves grid tests to utils tests).

PR #531, opened 2017-06-04
Grid.py Edits

MrDupin:
As per [these suggestions](https://github.com/aimacode/aima-python/issues/529#issuecomment-306014790).

I used `xA, yA = a` and `xB, yB = b` which I think looks better, but I could change it if requested. I also used capital letters for the constants, to comply with PEP8.

### DIFF
--- a/grid.py
+++ b/grid.py
@@ -6,7 +6,8 @@ import math
 
 from utils import clip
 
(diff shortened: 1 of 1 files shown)


norvig:
Looks good!

PR #535, opened 2017-06-06
Implemented fol_fc_ask()

Chipe1:
It is the inefficient implementation so I have commented out the test cases which takes 1 minute to run.
Also added `constant_symbols()` and its test.

### DIFF
--- a/logic.py
+++ b/logic.py
@@ -224,6 +224,16 @@ def prop_symbols(x):
         return list(set(symbol for arg in x.args for symbol in prop_symbols(arg)))
 
 
+def constant_symbols(x):
+    """Return a list of all constant symbols in x."""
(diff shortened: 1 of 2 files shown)


norvig:
This looks good.
Can we have a simpler test that runs faster?
Also, I note that we define knowledge bases with
     map(expr, ['A', 'B', 'C', ...])
could we just do
    expr('A, B, C, ...')
and since `expr` calls `eval`, wouldn't this just return a tuple of sub-exprs?

Chipe1:
I was unable to get the second format to work in multiple lines. `eval` throws `SyntaxError` due to the new line characters. I tried to use ''' inside """ but that didn't work either. I think it is better to leave it as it is. Using techniques to ignore the newline character would only make it harder to read.

Issue #515, opened 2017-05-04
FIFIOQueue, inefficient, unpythonic.


theelous3:
Hi there. I don't have the book that this codebase relates to, nor could I find a free version of it online. I came across this whilst helping someone and was just generally confused by the `FIFIOQueue` implementation.

I gather that the purpose of this codebase is not to write perfect pythonic code, so saying to throw the class away and just use a `deque` probably isn't helpful; however this brings up my fist issue / question. 

1. If the purpose of this code is to teach and / or give examples of how to implement data structures, the `FIFOQueue` being a direct wrapper of `collections.deque` doesn't accomplish that goal. If the code is going to be a wrapper around a `deque`, why not just subclass it?

2. Raising `Exception` in `pop` is unhelpful and redundant. `deque` already throws the correct exception. There are other general exceptions thrown in `extend` and `append`, which are unhelpful too.

3. Are the order of the args to __init__ intentional? They are backwards to deque, and python's style generally.

In light of this, I ran some timeit's on the current `FIFIOQueue` and a new `NewFiFoQueue` with a million calls to `.pop()`

```python
class FIFOQueue(Queue):
    def __init__(self, maxlen=None, items=[]):
        self.queue = collections.deque(items, maxlen)
    def pop(self):
        if len(self.queue) > 0:
            return self.queue.popleft()
        else :
            raise Exception('FIFOQueue is empty')

>>> timeit.timeit('[q.pop() for _ in range(20)]', setup='q = FIFOQueue(20000000, list(range(20000000)))', number=1000000, globals=globals())
6.41128579096403
```

```python
class NewFiFoQueue(collections.deque, Queue):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
    def pop(self):
        return self.popleft()

>>> timeit.timeit('[q.pop() for _ in range(20)]', setup='q = NewFiFoQueue(list(range(20000000)), maxlen=20000000)', number=1000000, globals=globals())
3.675234421971254
```

Directly subclassing `deque` results in code that runs in ~57% of the time taken by the current class.
Deque is pure C, in Cpython, meaning layering python on top of it hampers performance significantly.

(Just actually using a deque by using `appendleft` and `pop` to implement FiFo is even faster still, running in 2 second flat.)

Anyway, I'm more curious than anything. Thanks for your time.

Mark.

darius:
I think you're right that this class could be deleted. It was meant for Python 2.x where x is very small.

[...]

PR #544, opened 2017-06-09
Queues

MrDupin:
Note that this is an "experimental" PR. I made some changes according to #515 which are open to feedback, since I'm not entirely sure of the use of the classes and why things were built this way in the first place.

* I removed the abstract class `Queue`. Unless it plays a role in the pseudocode (or I'm missing something), I think it should be removed.

* Subclassed all the utilities in `FIFOQueue` from `collections.deque`, except `pop` since in `deque` it pops the rightmost element while we want to pop the leftmost one.

* Added `extend` to `PriorityQueue`, a function it was inheriting from `Queue`.

I would like some feedback on this, since I'm not entirely sure this is what was asked.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -651,31 +651,7 @@ class hashabledict(dict):
 # ______________________________________________________________________________
 # Queues: Stack, FIFOQueue, PriorityQueue
 
-# TODO: queue.PriorityQueue
 # TODO: Priority queues may not belong here -- see treatment in search.py
-
-
-class Queue:
-
-    """Queue is an abstract class/interface. There are three types:
-        Stack(): A Last In First Out Queue.
-        FIFOQueue(): A First In First Out Queue.
-        PriorityQueue(order, f): Queue in sorted order (default min-first).
-    Each type supports the following methods and functions:
-        q.append(item)  -- add an item to the queue
-        q.extend(items) -- equivalent to: for item in items: q.append(item)
-        q.pop()         -- return the top item from the queue
-        len(q)          -- number of items in q (also q.__len())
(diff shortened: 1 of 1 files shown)


norvig:
I'm not sure either. In search4e.ipynb, I have FrontierQ and FrontierPQ. But that too is "experimental".

MrDupin:
Hmm, looking at search4e.ipynb, I think we can go one of two ways:

a) Go a bit "hands-on", like in search4e, and don't subclass all the code from another class. This might be unpythonic and inefficient though, but the code is clearer and the student can pretty much read it all in one place.

b) Subclass as much as possible, like the edit I have above. The downside is that the student won't know the details of how the queues work, unless they read on `collections.deque` online.

In the end it comes down to what the student will need. Personally I don't feel a common data-structure is something students will pay much attention to, so we should go with my edit or something similar to it.

Another idea I had is to keep the class `Queue`, with the queues inheriting it. That way we can check if a queue is an instance of this Class (with `isinstance(q, Queue)`). It is not wildly important, but it could come up. Personally I don't feel though this is the route we should go, instead going with a bare-bones approach.

MrDupin:
Re-added the `Queue` class so that students can subclass it to their needs.

So right now the only changes I make are to update the queue functionality to Python 3.x. I think this is the safest way to go right now. If we need something different, we can always revisit this.

MrDupin:
Made another change. I added functions for `append`, `extend` etc. so that students can more easily subclass `FIFOQueue`.

I believe it can get merged now without issues, since the only major difference from the current code is performance.

MrDupin:
Making new PR.

PR #558, opened 2017-06-22
Added canvas_fol_bc_ask()

Chipe1:
Draws the proof tree constructed by backward chaining

### DIFF
--- a/canvas.py
+++ b/canvas.py
@@ -1,6 +1,7 @@
 from IPython.display import HTML, display
 from utils import argmax, argmin
 from games import TicTacToe, alphabeta_player, random_player, Fig52Extended, infinity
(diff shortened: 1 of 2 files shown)


Chipe1:
I would like to implement something similar for `fol_fc_ask()`, but I can't think of a way to present it without it looking similar to `fol_bc_ask()`. Any ideas?

norvig:
The completed proof will be the same, whether found through forward chaining or backward chaining. The difference will show up in an animation of the order in which the nodes are searched.

Chipe1:
@norvig Currently it only builds and displays the proof tree without any animation. Should I change it to display the nodes in a particular order? I don't think the animation will aid in understanding the workflow.

PR #560, opened 2017-06-24
Agent Tests

MrDupin:
Added tests for `RandomVacuumAgent` and `compare_agents` (the last one comes from #395, after I edited grammar/spelling).

We can now close #395. Thanks @articuno12 for the original test.

NOTE: I did not have much time to research this, but from the tests I ran it seems that `RandomVacuumAgent` always leaves the environment in the same state. If that is not the case and the agent leaves the environment in a random state, let me know and I will remove/update the test.

EDIT: Set `random.seed`, so it will now pass each time.

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -1,7 +1,11 @@
+import random
 from agents import Direction
 from agents import Agent
 from agents import ReflexVacuumAgent, ModelBasedVacuumAgent, TrivialVacuumEnvironment, compare_agents,\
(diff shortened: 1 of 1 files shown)


MrDupin:
UPDATE: Set `random.seed`, so that it doesn't matter if the function is deterministic or not. It will now pass every time.

norvig:
If there is no dirt in an environment, then anything an agent does, even a random one, will not add dirt.

PR #564, opened 2017-06-27
Rename canvas to notebook

MrDupin:
I renamed `canvas.py` to `notebook.py`. For the notebooks we sometimes need to run supplementary code for visualization which does not need to be shown. From now on we can simply add code to `notebook.py` and import it to the notebooks.

We can, alternatively, have a separate file in addition to `canvas.py`, but I believe this is better since it keeps it as simple as possible.


norvig:
This rename makes sense.

PR #573, opened 2017-07-03
N-gram Text Models

MrDupin:
There was a bug with the generation of sentences, where the first word of the input dataset would always be the first word in the sentence. Also, in my opinion our implementation of the `samples` function was not easy to read.

I rewrote part of the n-gram word model to fix the above. Now we have easy to understand code that works great, and we can have little nuggets of wisdom such as this: "son of a thousand other kindred subjects came soon to" (sentence for n=2).

Finally, I expanded the relevant section in the text notebook.

### DIFF
--- a/tests/test_text.py
+++ b/tests/test_text.py
@@ -6,6 +6,7 @@ from text import *
 from utils import isclose, open_data
 
 
+
 def test_text_models():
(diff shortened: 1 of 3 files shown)


norvig:
This is nice, Good changes.

PR #574, opened 2017-07-03
Learning, CSP and Games Notebook Headers

MrDupin:
Capitalized headers on the aforementioned notebooks to adhere to the uniform appearance.


Chipe1:
The canvas output for games.ipynb which was removed earlier is being readded in this PR. I believe it is a mistake.

MrDupin:
Good catch, but I don't think it really matters since in github the output doesn't take much space (a couple of scrollbars is not much).

PR #578, opened 2017-07-08
Learning

Chipe1:
Added section on Decision tree learner to notrbook
Added random forest code which is to be added in the 4th edition


MrDupin:
Nice addition. Is it possible to have some tests up?

Also, can you capitalize the header for the notebook section? Also, it seems that the Naive Bayes Classifier section has some issues in the "Examples" sub-section. You seem to have switched it from "Code" to "Markdown". Can you fix it?

norvig:
This is looking very nice. But maybe the notebook is getting too long -- about 25 pages now -- maybe break out the neural part to a separate notebook.

Chipe1:
@norvig Will the 4th edition go more in-depth with CNN, autoencoder etc. If so, should we use external libraries like TensorFlow in the separate notebook?

PR #581, opened 2017-07-09
Minor fixes

Chipe1:
Changes suggested in #578


MrDupin:
I noticed a couple of typos:

In your comment "A ensemble of Decision trees trained using bagging and feature bagging.", it should be 'An ensemble' and 'trees' should be capitalized.

Chipe1:
Fixed

PR #583, opened 2017-07-09
NLP Notebook: Add Question Asnwering

MrDupin:
Added a section for question answering. Even though it is a small section I believe it is an interesting concept worth noting.


norvig:
I agree this is interesting -- well noted.

PR #586, opened 2017-07-16
NLP Notebook: Languages

MrDupin:
Added a new section for languages in `nlp.ipynb`. I gave a quick overview of context-free grammars and probabilistic cfgs, then I gave an overview of lexicons and grammars and finally gave examples using our implementations.

* Work on notebook, including an image for tree parsing.

* Update `test_nlp.py`; added test for `Grammar`.

* Update `nlp.py` comments.

EDIT: I added `generate_random`. Even though most of the generated sentences are pretty much gibberish and can be very long, I think it is interesting to see what kind of phrases a very simplistic grammar can generate.

PS: So far in the chapter I have only encountered CFGs (and PCFGs). So even though there are other types of languages, I have refrained from writing on them since the book seems to focus on CFGs. If that is not the case and later on I encounter other grammars, I will amend the relevant section to add the rest of the grammars.


norvig:
For now, just CFG and PCFG is good.

MrDupin:
@norvig: OK great.

PR #587, opened 2017-07-18
Truncated SVD

Chipe1:
Added `norm()` and `truncated_svd()`
I'm not sure where to place the SVD code so I put it in utils.py


norvig:
This is good, but truncated_svd should probably be in learning.py, not utils.py

PR #593, opened 2017-07-23
Implementation: Current Best Learning

MrDupin:
Implemented the pseudocode from chapter 19, and added testcases.

Since the `learning` module is getting a bit too large, I decided to create a new module (`knowledge` for the chapter title "Knowledge in Learning"). I feel like this is the best way to go, but I can change it if this is not appropriate.

PS: This is an updated #591. Now performance is better + I settled on the format of the hypothesis (list of dictionaries, each dictionary represents a disjunction). I have worked on some other algorithms from the chapter, and this is the most appropriate format.


MrDupin:
Closing and reopening because an unrelated test failed.

PR #594, opened 2017-07-26
Knowledge Notebook: Current-Best Learning

MrDupin:
Added a section for the Current-Best Learning, including the two examples used in the tests. Also, fixed a small bug with the implementation.


norvig:
This is very nice. One thing: I don't like that each restaurant example takes up multiple rows of text and is hard to read.  I would define a function `rrow` (for "restaurant row", or pick a better name) and then write the table like this:

     restaurant = [
                            rrow('Yes No No Yes Some $$$ No Yes French 0-10 Yes'),
                            ...
                          ]

PR #595, opened 2017-07-27
Utils: Queues

MrDupin:
Similar to my last PR on the subject. As pointed out by #515, `FIFOQueue` can be improved. Currently it acts as a wrapper to `collections.deque`, but I feel it is better if it directly subclassed it and then let the students decide if they want to further subclass that to suit their needs.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -689,7 +689,7 @@ def Stack():
     return []
 
 
-class FIFOQueue(Queue):
+class FIFOQueue(Queue, collections.deque):
 
     """A First-In-First-Out Queue."""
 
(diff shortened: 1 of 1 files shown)


darius:
Add some tests? It looks incorrect at first glance: `self.queue` doesn't appear to be touched by the append and extend methods, and the `items` parameter gets a shared mutable default value.

MrDupin:
Tests for `FIFOQueue` were there before this PR ([link](https://github.com/aimacode/aima-python/blob/master/tests/test_utils.py)). I didn't change them at all. Also, the class is used in `search` and its tests.

I probably shouldn't have left the init method in as is. I left the `self.queue` initialization in case students want to further subclass `FIFOQueue` (so that they know they can work with `queue`). As it stands, the code doesn't use `queue`. It directly subclasses `collections.deque`, instead of wrapping it.

Maybe I can replace the init code with `pass`? I think that would be better.

MrDupin:
On second thought, I don't think this PR is necessary. The impact on performance in negligent in the examples/problems we want to showcase and the original code is at least as clean as the current one. Closing this.

PR #602, opened 2017-08-03
Added Monte Carlo localization

Chipe1:
Added test with a discrete space


norvig:
This looks good! One small thing: instead of `[i, j]` for a point, let's use `(i, j)`.

Chipe1:
Changes in #603

PR #612, opened 2017-08-10
NLP: Chart Parsing

MrDupin:
Added a test and a section to the notebook for chart parsing. Also, updated the docstring in `nlp.py` to the 3rd edition.


norvig:
This looks nice. One style comment: you have:

```
Lexicon(
    Verb="is | say | are", ...
```

and I know that is the correct PEP 8 style for keyword args, but seen as a grammar rather than a function call, I think it would look prettier as:

```
Lexicon(
    Verb = "is | say | are", ...
```

Issue #613, opened 2017-08-10
%psource


norvig:
We use %psource in notebooks to refer to code, and that is good. But the code is displayed in a popup window, and one drawback is that you can only display one function at a time this way. Another drawback is that the code doesn't scroll with the rst of the document (this can sometimes be an advantage, sometimes a drawback).

I propose to add to utils the following function, and prefer to use it over %psource:

```
import inspect

def psource(*functions):
    "Print the source code for the given function(s)."
    print('\n\n'.join(inspect.getsource(fn) for fn in functions))
```

MrDupin:
PR #614 takes care of this. I decided it was better to add the function to `notebook` instead of `utils`, to keep utils for implementation only stuff.

VishnuDuttSharma:
@MrDupin I want to add a beautified version of the commit you made. Can you help me out with the process?

MrDupin:
Sure, how can I help?

On the code, I would suggest you join on two newlines ( `'\n\n'.join ...` ) instead of none. It looks better to have the functions separated.

My only concern is whether or not `pygments` ships with Python or if the user will have to install it themselves. We try to keep the repository as free from external libraries as possible.

If you need anything else, feel free to ask.

VishnuDuttSharma:
I think it should not be a problem, since this is required as a feature for notebooks and notebooks are generally installed using anaconda. 
However, to be on safe side, should I put it in a try-except block, depending upon availability of `pygment`?

MrDupin:
I like this idea, yes. If pygments is installed, use that, otherwise use the default one.

I think this very useful since syntax highlighting is very important for a lot of people, so I suggest you make a PR after your changes and let Mr. Norvig decide. It looks good to me though.

MrDupin:
Since the PR got merged, I believe this can be closed.

norvig:
Thanks for the good work on this!

PR #614, opened 2017-08-11
Notebook: NLP Styling + psource

MrDupin:
Added `psource` to `notebook.py` (I think it's better to add it to `notebook` than to `utils`) as per #613. Also, made style changes to lexicon/grammar calls.


norvig:
You're right, `notebook` is a better place for it; thanks.

PR #616, opened 2017-08-12
Pseudocode In Notebooks

MrDupin:
I properly packaged pseudocode printing. The code can now be found in `notebook.py`. Also, previously I didn't like very much how we would also print the name of the algorithm (the pseudocode is already in a section titled after the algorithm, so it's a bit overboard, especially since it is formatted in the largest header), so I made a couple of adjustments to amend that.

Now we only show the appropriate parts of the pseudocode, including the version of the book.

I have made the corresponding changed to the knowledge notebook, for a sample.

### DIFF
--- a/knowledge.ipynb
+++ b/knowledge.ipynb
@@ -19,7 +19,9 @@
    },
    "outputs": [],
    "source": [
-    "from knowledge import *"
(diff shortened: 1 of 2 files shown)


norvig:
nice!

PR #619, opened 2017-08-13
Python implementation for psource (#613)

VishnuDuttSharma:
Added the python implementation for `psource` to get the output in a cell instead of a pop-up to `notebook.py` as suggested by @MrDupin . Also added syntax highlighting to this code, with dependence upon the availability of the module `pygments`.


MrDupin:
There have been some conflicts since you last edited `notebook`, may you resolve them?

PR #625, opened 2017-08-17
FOIL

Chipe1:
Added a basic implementation of FOIL.
The original algo, being intended for use in Prolog, is not directly compatible with python. So I've used FolKB to store and process the knowledge. This works for easy examples like the network given in the original paper and the family tree example.
The removing of positive examples is not recursive, and this is leading to technically correct but long and overfitted clauses. Any help in fixing that would be appreciated.

### DIFF
--- a/knowledge.py
+++ b/knowledge.py
@@ -1,9 +1,12 @@
 """Knowledge in learning, Chapter 19"""
 
 from random import shuffle
+from math import log
 from utils import powerset
 from collections import defaultdict
-from itertools import combinations
+from itertools import combinations, product
+from logic import (FolKB, constant_symbols, predicate_symbols, standardize_variables,
+                   variables, is_definite_clause, subst, expr, Expr)
 
 # ______________________________________________________________________________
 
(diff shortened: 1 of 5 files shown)


knowledge.py
```diff
@@ -231,6 +234,117 @@ def consistent_det(A, E):
 # ______________________________________________________________________________
 
 
+class FOIL_container(FolKB):
```
norvig: I think this should be called FOILKB.

logic.py
```diff
@@ -216,23 +216,33 @@ def tt_check_all(kb, alpha, symbols, model):
...
+    """Return the set of all propositional symbols in x."""
     if not isinstance(x, Expr):
-        return []
+        return set()
     elif is_prop_symbol(x.op):
-        return [x]
+        return {x}
     else:
-        return list(set(symbol for arg in x.args for symbol in prop_symbols(arg)))
+        return {symbol for arg in x.args for symbol in prop_symbols(arg)}
```
norvig: OK, or could be
`return set().union(map(prop_symbols, x.args))`

logic.py
```diff
@@ -216,23 +216,33 @@ def tt_check_all(kb, alpha, symbols, model):
...
+    """Return the set of all constant symbols in x."""
     if not isinstance(x, Expr):
-        return []
+        return set()
     elif is_prop_symbol(x.op) and not x.args:
-        return [x]
+        return {x}
     else:
-        return list({symbol for arg in x.args for symbol in constant_symbols(arg)})
+        return {symbol for arg in x.args for symbol in constant_symbols(arg)}
```
norvig: Similar here as for prop_symbols

PR #626, opened 2017-08-18
Added Append

Chipe1:
Also added the APPEND_container class to implement variable binding and stack/trail.
I have not included stack points and as a result unify only returns one possible solution at a time.


norvig:
I don't understand the purpose of this? We already have variable binding lists; what is this mechanism for? Can you give a description and some examples of use?

Chipe1:
The binding dictionary is similar to substitution dictionary for `Expr` variables. We could use a single dictionary but it would need to be global (or pass it around as an additional argument) for functions like reset-trail to work. We also need a separate unify function which can handle list of variables. I made the dictionary an attribute, and unify a method of the class instead.
Usage example:
`env = APPEND_container()`
`env.add_var(x, True)`
This would make `env[x]` return `True`.
Most of the other methods like `new_variable()` and `reset_trail()` would only be used internally. Their usage can be found in the tests.

PR #634, opened 2017-08-25
Minor changes

Chipe1:
Based on the comments in #625

### DIFF
--- a/logic.py
+++ b/logic.py
@@ -222,7 +222,7 @@ def prop_symbols(x):
     elif is_prop_symbol(x.op):
         return {x}
     else:
-        return {symbol for arg in x.args for symbol in prop_symbols(arg)}
+        return set().union(*map(prop_symbols, x.args))
 
 
 def constant_symbols(x):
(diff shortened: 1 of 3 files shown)


logic.py
```diff
@@ -222,7 +222,7 @@ def prop_symbols(x):
     elif is_prop_symbol(x.op):
         return {x}
     else:
-        return {symbol for arg in x.args for symbol in prop_symbols(arg)}
+        return set().union(*map(prop_symbols, x.args))
```
MrDupin: To be honest, I prefer the comprehension to the set union with the map pointer, since it is much easier to understand (at least for me).

If both are correct, I would change it to the comprehension (which is simpler and stylistically closer to the way we do things in the repository).

That is unless the union is much, much faster or is the most popular way to deal with the problem.

logic.py
```diff
@@ -232,7 +232,7 @@ def constant_symbols(x):
     elif is_prop_symbol(x.op) and not x.args:
         return {x}
     else:
-        return {symbol for arg in x.args for symbol in constant_symbols(arg)}
+        return set().union(*map(constant_symbols, x.args))
```
MrDupin: Same as my above review.

logic.py
Chipe1: The map method is efficient but I dont think we need to worry about speed as we are using only small expressions. Since it is not directly mentioned in any pseudocode, people could look at it like a simple blackbox function. Finally it's upto @norvig 
Maybe the we can use both. One in each of the functions :D

logic.py
norvig: If you think the comprehension is easier to understand, let's stick with that.

PR #637, opened 2017-08-27
Fix text.py - aima-data links

MrDupin:
Some links to aima-data are now broken, this PR fixes them, in conjunction with [this aima-data PR](https://github.com/aimacode/aima-data/pull/11).

Also, the repository link to aima-data is a bit old (branch 6ce56c0). Is it possible to update it? I don't think it will make any difference, but if it's not much trouble it might be worth it.

### DIFF
--- a/nlp_apps.ipynb
+++ b/nlp_apps.ipynb
@@ -49,7 +49,7 @@
     "\n",
     "P_flatland = NgramCharModel(2, wordseq)\n",
     "\n",
-    "faust = open_data(\"faust.txt\").read()\n",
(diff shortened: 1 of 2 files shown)


norvig:
I'm not sure what link / where you want to update?

MrDupin:
I meant to say the [aima-data submodule](http://imgur.com/NElSrLs). It links to [this](https://github.com/aimacode/aima-data/tree/6ce56c0b67206bae91b04fb20f0d8d70c9a86b6a), which is outdated.

I have opened an [issue](https://github.com/aimacode/aima-python/issues/638) so this doesn't get lost.

PR #647, opened 2017-09-09
Updating Submodule

MrDupin:
Updated the aima-data submodule. Also wrote a short guide on how to update the submodule, in case this needs to be done and I'm not around.


norvig:
Very helpful! I wish it wasn't so complicated, but great to have this written down.

greysteil:
@jfitie - would you be up for automating submodule updates? I just added support for it to [Dependabot](https://dependabot.com). Totally free service - just want to get people using it :)

PR #653, opened 2017-10-23
Probability: Notebook + gibbs_ask test

MrDupin:
I showed the outputs on `probability.ipynb`, which were previously hidden, and I added a test for `gibbs_ask`.


norvig:
Thanks!

PR #655, opened 2017-10-25
Fixed a syntax error

A-M-Coder:
I changed two statements if and else which were the reason of syntax error


norvig:
No, the code is correct as it is. Your problem seems to be that you are trying to run the code in Python 2, when Python 3 is required (or for this particular issue, at least `from __future__ import print_function`).

A-M-Coder:
Thank you for reviewing my pull request and guiding me.

PR #684, opened 2018-01-10
 Update PeakFindingProblem code to allow diagonal motion

AdityaDaflapurkar:
Completed TODO: Add flag to allow diagonal motion in PeakFindingProblem class in search.py


norvig:
Too many `if / else` statements. Better to do this with a dict of actions, of the form `dict(NW=(-1, 1), ...)`, where there is one dict with 4 entries, and another with 8, and users have the option to pass in a different dict (such as knight moves).

AdityaDaflapurkar:
@norvig : I have updated the code according to your instructions. However, I have removed the allow_diagonal_motion flag parameter and instead used just the defined_actions dict parameter(because allow_diagonal_motion was unused for non-default dict values). The two default values of the dict for 4 directions and 8 directions are included as variables in search.py. Please review and suggest further changes if any.

search.py
```diff
@@ -509,38 +509,40 @@ def and_search(states, problem, path):
     # body of and or search
     return or_search(problem.initial, problem, [])
 
+# Pre-defined actions for PeakFindingProblem
+primary_actions = { 'W':(-1,0), 'N':(0,1), 'E':(1,0), 'S':(0,-1) }
```
norvig: - Better names would be something like `directions4` and `directions8`.
- There should be a space after a comma.
- do `directions8 = directions4 | {'NW': ... }

search.py
```diff
@@ -509,38 +509,40 @@ def and_search(states, problem, path):
...
+aggregate_actions = { 'W':(-1,0), 'N':(0,1), 'E':(1,0), 'S':(0,-1), 'NW':(-1,1), 'NE':(1,1), 'SE':(1,-1), 'SW':(-1,-1) }
 
 class PeakFindingProblem(Problem):
     """Problem of finding the highest peak in a limited grid"""
 
-    def __init__(self, initial, grid):
+    def __init__(self, initial, grid, defined_actions = primary_actions):
```
norvig: No space around `=` in optional arguments: `defined_actions=primary_actions`

search.py
```diff
@@ -509,38 +509,40 @@ def and_search(states, problem, path):
...
+        for action in self.defined_actions:
+            next_state = (state[0] + self.defined_actions[action][0], state[1] + self.defined_actions[action][1]) 
+            if next_state[0] >= 0 and next_state[1] >= 0 and next_state[0] <= self.n - 1 and next_state[1] <= self.m - 1:
+                allowed_actions.append(action)
+
         return allowed_actions
 
     def result(self, state, action):
         """Moves in the direction specified by action"""
         x, y = state
```
norvig: I think we have defined an `add` function for points/vectors. I forget what it is called, but if it is `add` then `result` should be just `return add(state, action)`

search.py
AdityaDaflapurkar: The `|` operator cannot be used to append dicts. Should I do it in the following way:
directions8 = dict(directions4)
directions8.update({ 'NW': ... })

Issue #688, opened 2018-01-12
Error loading search.ipynb


apb7:
On opening search.ipynb, the following error occurs:  

![screenshot-localhost-8888-2018-01-11-10-30-03-094](https://user-images.githubusercontent.com/24826041/34879544-9a03d4e2-f7d3-11e7-9fd1-912a62f7e9fb.png)

nvinayvarma189:
I think somebody has got merge conflicts in the file - those <<< and === and >>> lines are how git shows its merge conflicts. To make it valid, clearing the merge conflicts.


```
<<<<<< HEAD
   Either keep this line...
=======
   ... or this one. But not both!
>>>>>>> a1f5acbaa00bc1b1e5fc3532ba19a013a271542c
```

and also remove the lines having  '<<<<<< HEAD',  '=======',  '>>>>>>> a1f5acbaa00bc1b1e5fc3532ba19a013a271542c'. Those are the git conflict makers.

and then [nbdime](http://nbdime.readthedocs.io/en/stable/) can be used to avoid git doing that again. I hope this solves the issue.

apb7:
Hello  @nvinayvarma189! Thanks for the help! But I think this issue has been resolved with PR #689 (though the PR hasn't been merged yet). 
Thanks!

nvinayvarma189:
Oh sorry, I did not look at it. Thanks for letting me know.

nvinayvarma189:
Hey @apb7 I find the PR for this issue is still not merged. I see all the checks have passed but, is there any problem still not solved ?

apb7:
Hi @nvinayvarma189! The PR will be merged by @norvig.  
Thanks!

PR #689, opened 2018-01-12
Solve Issue of Loading search.ipynb

Dragneel7:
#688

### DIFF
--- a/search.ipynb
+++ b/search.ipynb
@@ -1268,7 +1268,7 @@
     "              | 5 | 0 | 6 |                       | 3 | 4 | 5 |\n",
     "              | 8 | 3 | 1 |                       | 6 | 7 | 8 |\n",
     "              \n",
-    "We have a total of 9 blank tiles giving us a total of 9! initial configuration but not all of these are solvable, the solvability of a configuration can be checked by calculating the Inversion Permutation.\n",
+    "We have a total of 9 blank tiles giving us a total of 9! initial configuration but not all of these are solvable, the solvability of a configuration can be checked by calculating the Inversion Permutation. If the total Inversion Permutation is even then the initial configuration is solvable else not.\n",
     "\n",
     "#### Heuristics :-\n",
     "\n",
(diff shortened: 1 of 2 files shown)


Dragneel7:
@MrDupin I have solved this issue, I just have to rebase it with master and then it is ready to go.
@varunvora Are you working on this? I started working on it as no one had claimed the issue and the errors were because of conflicts in mine and abp7's code.

Dragneel7:
@norvig @MrDupin Please review the PR. I have fixed the issue and cleaned the code in accordance to the coding trend of the repository  with the algorithm in  "search.py" and only the heuristics and results in the notebook.
@apb7 Kindly check if the issue was resolved.
Thanks.

MrDupin:
I believe the issue is resolved here. Nice work!

apb7:
Hello @norvig and @MrDupin! The issue seems to be resolved. Just a little doubt, @Dragneel7, shouldn't we remove the '0' tile as the 8 puzzle problem has 8 tiles and a space to move the tiles (which is in conformity with the problem stated in the book). Also, this has been mentioned in the notebook as well:  
>The 8 Puzzle Problem consists of a 3x3 tray in which the goal is to get the initial configuration to the goal state by shifting the numbered tiles into the blank space. 

Please feel free to correct me!  
Thanks!

Dragneel7:
@apb7 That is true, I had intended to mention this in the description of the program to treat the tile 0 as blank. I guess I forgot to do the same.
Still with the example diagram it is pretty clear what is happening, moreover we are concerned not with solving the problem but with the heuristics that solve it, So i guess '0' works fine too.
Thuough it won't be any trouble to change it.

MrDupin:
I noticed a little typo, if you can get it fixed. In the "A* Heuristics" chapter, in the first sentence you write:

"Different heuristics provide different have different efficiency"

Where you write both "provide different" and "have different". You should choose one or the other (I think the first sounds better).

Dragneel7:
@norvig The python notebooks are not getting rendered, it's maybe due to the issue that is being solved by this PR. Can you please look into the matter.
Thanks

Issue #659, opened 2017-11-18
Tkinter Applications


MrDupin:
@apb7 made a tic-tac-toe game using tkinter and it was really cool. I would love to see that idea expanded to more algorithms. The only additional requirement is `python3-tk`, and the new scripts can be added to a separate folder to keep the main directory as uncluttered as possible.

@norvig, what do you think? I personally feel this is a nice addition to the notebooks, especially since it gives us more freedom. (there is also the possibility of adding tkinter code to notebooks and executing it from there)

apb7:
@MrDupin: I'll be implementing other algorithms as well. And as @Chipe1 suggested, an additional side panel to illustrate the working of the respective algorithm can also be added. 
The directory structure: A separate `gui` folder to place all `Tkinter` files.

apb7:
@MrDupin: Sorry for catching up so late! Shall I make a PR for the tic-tac-toe game? Also which other algorithms shall I expand this idea to now?

MrDupin:
No problem. Yes, a PR sounds fine. I suggest you make a PR for the tic-tac-toe, and if it gets accepted you can work on other stuff.

For example, the search algorithms can be implemented in a guy. The nodes can be some random boxes, and with a 'Next' button the next move of the algorithm is shown, maybe by using a color scheme.

Apart from that, you can look at what you can implement from the notebooks into a visual way.

apb7:
Okay! Thanks, @MrDupin. I'll get to work and keep you posted.

apb7:
@MrDupin and @norvig : I have worked out a bit on the GUI for Searching and have pushed the changes in my version of the [repository](https://github.com/apb7/aima-python/) under `gui`. A map of Romania has been added as depicted in the book and notebook. Various search techniques will be added soon.  
Please feel free to suggest any improvements!

MrDupin:
The map looks good. The frame of the window though is a bit too small. The "Currently Exploring" item on the legend is cut off.

apb7:
I'll make the necessary changes tonight. Also, I have plotted the actual coordinates as given in the `romania_map` due to which the windows frame seems to be small. Therefore I'll chose some other representative coordinates for this purpose. Thanks!

apb7:
@MrDupin and @norvig: The following updates have been made:  
* Breadth-first Tree Search has been added.  
* Options for `start` city and `goal` city provided.  
* Window size has been adjusted to fit the `canvas` widget.  
[...]

PR #693, opened 2018-01-15
Adding Tkinter GUI

apb7:
* Issue #659   
* Added `vacuum_agent.py` and `xy_vacuum_environment.py`. 
* `vacuum_agent.py` creates a GUI for the two-tiles vacuum environment.  
* `xy_vacuum_environment.py` creates a GUI for the two-dimensional dynamic vacuum environment which       can be changed at any point of time by the user.

### DIFF
--- a/gui/vacuum_agent.py
+++ b/gui/vacuum_agent.py
@@ -14,7 +14,7 @@ class Gui(Environment):
     or Clean. The agent perceives its location and the location's
     status."""
 
-    def __init__(self, root, height=300, width=380, count=1):
+    def __init__(self, root, height=300, width=380):
         super().__init__()
         self.status = {loc_A: 'Clean',
                        loc_B: 'Clean'}
@@ -23,7 +23,6 @@ class Gui(Environment):
         self.width = width
         self.canvas = None
         self.buttons = []
-        self.count = count
         self.create_canvas()
         self.create_buttons()
 
@@ -48,9 +47,9 @@ class Gui(Environment):
         elif action == 'Suck':
             if self.status[agent.location] == 'Dirty':
                 if agent.location == loc_A:
-                    self.buttons[0].config(bg='white')
+                    self.buttons[0].config(bg='white', activebackground='light grey')
                 else:
-                    self.buttons[1].config(bg='white')
+                    self.buttons[1].config(bg='white', activebackground='light grey')
                 agent.performance += 10
             self.status[agent.location] = 'Clean'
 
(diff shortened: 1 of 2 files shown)


apb7:
Hello @norvig and @MrDupin! Can you please review this once?  
Thanks!

MrDupin:
What agent is the `xyEnvironment` implementing? It is not very useful, it seems to be going off in random directions, even when dirt is right next to the agent. Is it how it's supposed to work? Also, the 'click-to-change-tile' method is a bit clunky. Right now it rotates globally around W/D/Empty. It would be best if it was the individual tile was rotating around the options. I think this can be implemented by checking the tile value when the user clicks on it. If it is 'D', it becomes empty, if it is 'W' it becomes 'D' and if it is Empty it becomes 'W' (or something along these lines). I think that would make things easier.

In `vacuum_agent.py`, the scenes appears too low on the window. Is it possible to get it centered? Apart from that, I would like to see the hover-over color of the buttons changed. If I hover over a Dirt tile, it should become a dark brown, and if I hover over an Empty tile, it should become gray. Right now the button becomes gray no matter what its value is, and it is a bit confusing. Also, the left button doesn't seem to work when I press it.

apb7:
@MrDupin: Currently the `xyEnvironment` is using `random.choice` to generate its next action: If the percept is 'Bump', it has 50% probability to go right and 50% to go left. If the tile is clean, there is 25% probability to go right, 25% to go left and 75% to move forward (Similar to that of the `EnergeticBlindDog` given in the notebook): [XYReflexAgentProgram](https://github.com/apb7/aima-python/blob/gui/gui/xy_vacuum_environment.py#L127:L143). This can be changed whenever we plan to expand the idea!  
And yes! I also found the click-to-change-tile a bit clunky. I'll incorporate your idea there :)
In `vacuum_agent.py`, I can surely center the scene. Although I am not sure if the hover-color can be changed according to the button. Nevertheless, I'll look around for a way. And, the left button works for me though I'll check it once again!  
Thanks, for an in-depth review!

MrDupin:
Ah, got it. The `xy_environment` is fine then. Looking forward to seeing this expanded!

apb7:
Thank you @MrDupin! I'll make other changes soon.

apb7:
Hello @norvig and @MrDupin! I have made the changes. The hover-over-color now changes according to the block color: a light brown for dirty block and gray for a clean block. I have centered the scene in `vacuum_agent.py` and made sure that all the buttons work.  
For the `xy_environment`, click-to-change-tile works in the same way as suggested above.  
Please suggest further improvements!  
Thanks!

MrDupin:
I like the changes, nice work!

apb7:
Thank you @MrDupin!

Issue #695, opened 2018-01-20
Add hill climb search example to search.ipynb


Dragneel7:
The following algorithm has not been implemented in the search.ipynb notebook. We can add this example by using a self created or some well suited data set, visualize the data set with a line graph thus showng local and global maxima's.
Show the result visually after the algorithm is run.
@MrDupin Should I implement it?

MrDupin:
Sure, sounds good.

Dragneel7:
@norvig  I am planning to use the cipher example from your work to illustrate hill climb search, I hope it won't be a problem.

MrDupin:
@Dragneel7, sure, go ahead!

dmeoli:
`search.ipynb` already has a dedicated **HILL CLIMBING** section: `psource(hill_climbing)`, an explanation of local/global maxima, and a worked example running `hill_climbing` on a TSP instance (also reused in the Simulated Annealing section for a comparison). Closing as completed.

PR #697, opened 2018-01-24
add hill climb search

Dragneel7:
#695 Solve Caeser Cipher Problem Using Hill Climb Search.


Dragneel7:
@norvig @MrDupin Please review the PR.
Thanks.

norvig:
This is nice work, but I think it should be refactored. Your `CeaserCipher` class does three things: hillclimbing, applying hillclimbing to Caeser Cipher, and plotting the result. These should be broken out into three different functions, not one class. Make a general hillclimbing function that takes a starting point, a function to generate neighboring points in the space, and a function to give a value that is to be maximized. Then call it with appropriate arguments to solve the cipher, and plot results.

Dragneel7:
Thanks for the suggestion, I will try to refactor the code as mentioned.

Dragneel7:
@norvig , I have a doubt, the approach I have taken to solve the cipher is to store all the possible(26) combinations of the initial message in an array and then iterate over the array to find the most suitable solution.
I am stuck at the part as I cannot generate neighboring states and cause each element of the array is independent of the previous and hence no tree like structure is generated.
As of now I have removed the class and  refactored the code in a 3 function way. But I am a bit confused how to make a generalize hill climb for this specialized problem.
Thanks.

MrDupin:
I am not that familiar with the hill-climbing/caesar cipher combo, so I can't be of much help. Maybe you can receive as input the possible combinations and go from there?

I can see that in some problems that input would be large, but remember we are not building real-world applications, so don't worry about size. We assume the students will have small examples.

norvig:
`hill-climbing-search` should be  a separate function, like `uniform-cost-search`. It should take as inputs an initial state, a function to generate neighboring states, and an evaluation function to say how good each neighbor is. Neighbors could be formed by swapping two letters, so there are 26*25, not 26! options. Or you could choose a random subset of those. See page 232 of http://norvig.com/ngrams/ch14.pdf

Dragneel7:
Thanks for the help, I will implement it as soon as possible

Dragneel7:
I read the link, what you mentioned above is true for substitution cipher, but for the case of caeser cipher won't the sample space will be 26 ( 1 for each of the 26 keys).
Please correct me if I have misunderstood.

norvig:
That's right -- for Caeser (shift) cipher, there are just 26 possibilities (25 without the identity shift), and you can try them all. No hillclimbing required.

dmeoli:
Thanks @Dragneel7. hill_climbing already exists in search.py, this PR also creates a mis-named notebook (search-4e.ipynb vs the repo's search4e.ipynb), and a Caesar-cipher hill-climb demo is somewhat tangential (text.py already has cipher decoding). Closing as superseded/stale — thanks!

Issue #707, opened 2018-02-02
Add Best First Search example to search.ipynb


nvinayvarma189:
I would like to add the example to the search notebook

MrDupin:
Awesome! Don't forget to add the section to the notebook contents and you are golden.

PR #708, opened 2018-02-02
added Best First search in search.ipynb

nvinayvarma189:
@MrDupin  Can you please review this and kindly notify me if something is not correct. This is for issue #707


MrDupin:
Looks good to me (although I can't execute it to check if it works, I am in the middle of exams and I don't have JS Widget installed).

nvinayvarma189:
I tested it between different cities and It works completely fine when I run it.

PR #721, opened 2018-02-10
Add vacuum_world.ipynb

### LINKED ISSUE #720: Adding and Explaining Vacuum Agents
Hello @norvig and @MrDupin! Currently, we do not have an implementation and visualization of `vacuum agents` in `agents.ipynb`.   
I am planning to add and show the working of all the vacuum agents we have in `agents.py`: Random-Vacuum-Agent, Model-Based-Vacuum-Agent, Table-Driven-Vacuum-Agent, Table-Driven-Agent, Reflex-Vacuum-Agent, Simple-Reflex-Agent and Model-Based-Reflex-Agent. I will also focus on the structure of the agents as given in the book.  
I will begin with a simple two-state environment along with explaining the working of each one's `program` and will then work out on expanding it to a two-dimensional environment.  
Also, I think it would be better if we implement this complete topic in a separate notebook pertaining to vacuum agents itself (since the book follows this example): `vacuum_world.ipynb`.  
I believe that this will be a prominent part of the notebook since it is the "building block" for agents.  This notebook will also serve the purpose of an explanation for the agent module.
Please suggest!  
Thanks!

apb7:
This PR:  
* Fixes #720 
* Adds a new notebook: `vacuum_world.ipynb` which explains the various agent programs given in the book.  
* Also adds appropriate block diagrams to demonstrate each agent's mechanism, as described in the book. 
(I have created these schematic diagrams on my own with the help of the book so there is no copyright problem.)  

@norvig and @MrDupin: Please review the PR.  
Thanks!


apb7:
@norvig and @MrDupin: I will also add visualizations as well as a two-dimensional environment for the same.  
Thanks!

PR #727, opened 2018-02-13
Fixed TypeErrors (Fixed #711)

### LINKED ISSUE #711: NQueensProblem doesn't work with search methods
If I `import search` and then
```
puzzle = search.NQueensProblem(8)
print(search.breadth_first_search(puzzle))
```
I get:
```
TypeError: unhashable type: 'list'
```
In order to work with the various search algorithms, the states have to be tuples, not mutable lists.

ad71:
- typecasted unhashable lists into tuples wherever sets are used
- NQueensProblem is now compatible with `graph_search`, `depth_first_graph_search`, `breadth_first_search`, `best_first_graph_search`, `uniform_cost_search` and `astar_search`
- Closes #711

### DIFF
--- a/search.py
+++ b/search.py
@@ -202,9 +202,9 @@ def graph_search(problem, frontier):
         node = frontier.pop()
         if problem.goal_test(node.state):
             return node
-        explored.add(node.state)
+        explored.add(tuple(node.state))
         frontier.extend(child for child in node.expand(problem)
-                        if child.state not in explored and
+                        if tuple(child.state) not in explored and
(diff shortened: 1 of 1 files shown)


norvig:
No, that's not the right way top handle this. Calling `tuple(node.state)` is a good idea if the state is a list, but if it is a `str` this is wasteful, and if it is an `int` this is an error. Instead, the right thing is that problems should produce states that are hashable.

ad71:
I'm sorry I completely overlooked this aspect.

PR #739, opened 2018-02-14
Adaboost example

### LINKED ISSUE #728: Add examples and tests for AdaBoost algorithm
Hey @norvig, @MrDupin: currently, the `learning.ipynb` notebook lacks an example for **ensemble learning**. Since the `perceptron` classifier gives an accuracy of almost 68.8% (not a weak learner though) on the **iris** dataset, I was thinking that maybe we can use it to add an example for **AdaBoost algorithm** and show how it works. We can also add code for **decision stump**, but since its pseudo code has not been provided maybe we can go with the perceptron classifier?
I also wanted to add some tests for **AdaBoost**.

What do you think?

aswanipranjal:
This PR:

- Closes #728.
- Adds example for AdaBoost in `learning.ipynb`
- Adds tests for AdaBoost

@norvig, @MrDupin please review this PR.
Thank you!

### DIFF
--- a/learning.ipynb
+++ b/learning.ipynb
@@ -11,7 +11,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
    "metadata": {
     "collapsed": true
    },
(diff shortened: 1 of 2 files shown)


aswanipranjal:
@MrDupin, the second last test fails here, although it passes locally on my laptop. Should i change the tests to use `grade_learner` instead?

MrDupin:
Yes, use `grade_learner`. In Travis we cannot set `random.seed`, so the tests randomly fail. With `grade_learner` the chances of the test failing are reduced significantly.

norvig:
@MrDupin I don't understand what's up with `random.seed` in Travis? If you call `random.seed(42)` what happens?

MrDupin:
@norvig I am not entirely sure what happens either. There is a problem in the `learning` tests, where even though we set a random seed, the tests of non-deterministic algorithms (like the neural net) occasionally fail. Locally, we always get the same results, but through Travis it seems that is not the case.

PR #738, opened 2018-02-14
Add table for Table Driven Vacuum Agent in vacuum_world.ipynb

apb7:
This PR modifies the table for `TableDrivenVacuumAgent` in `vacuum_world.ipynb`.  

@norvig and @MrDupin: Please review this PR.  
Thanks!

PR #732, opened 2018-02-13
Fix #731: Add table and tests for TableDrivenVacuumAgent

apb7:
This PR:  
* Fixes #731 
* Implements the table for the `TableDrivenVacuumAgent` in the `TrivialVacuumEnvironment`
* Adds a test for `TableDrivenVacuumAgent` in the `TrivialVacuumEnvironment`

@norvig and @MrDupin: Please review this PR.  
Thanks!

PR #747, opened 2018-02-15
Add test for Table-Driven-Vacuum-Agent.

roberthoenig:
Note that this changes the percept sequence table in the code, it now slightly differs from table 2.3
in the book. The underlying problem here is that a full table can principally not be specified, so '...' is used to represent the missing table entries. While this works for the book, the code requires some additional table entries for the agent to succeed.

I can see how this approach might be considered less than optimal. If that's the case, I'm happy to discuss alternative solutions.


apb7:
Hello, @roberthoenig! This issue has already been fixed in PR#732 and #738. Please  see those PRs once.  
Thanks!

roberthoenig:
@apb7 thanks for the heads-up, I should've checked existing issues/PRs more carefully :) Closing this in favor of #732.

apb7:
No problem, @roberthoenig! Thank you for your cooperation. Looking forward to your contribution :)

PR #750, opened 2018-02-16
Fix various typos.

roberthoenig:


MrDupin:
Thanks for the help, nice catches!

A little request: Please thin out PRs that edit a lot of notebooks. When this gets merged, it's very likely that merge conflicts will arise. Merge conflicts on notebooks are not that simply to fix so it will require a lot of work from the rest.

Thanks!

roberthoenig:
Thanks for the review! Re merge conflicts: Sounds reasonable, I’ll watch out for notebooks in the future.  What’s the next step for this PR to get merged? The options I see are 
* leave it as it is.
* split into two commits, one for non-notebook and one for notebook, so the first one can easily get cherry picked.
* split into two PRs, so the first one can easily get merged.

MrDupin:
It is up to you, but I would leave it as is. If in the future you see a PR already editing a notebook with a typo, you can message the PR creator and let them know of the typo so that they can fix it.

PR #751, opened 2018-02-16
Change ReflexVaccumAgent. Add class Rule.

kushalchauhan98:
Change ReflexVaccumAgent. Add class Rule.

This PR:
- Implements ReflexVaccumAgent using SimpleReflexAgentProgram() and removes redundant implementation of program() in ReflexVaccumAgent
- Adds Rule class to facilitate the same
- Inherently adds test for SimpleReplexAgent

(As the Reflex Vaccum Agent is a special case of the generic Simple Reflex Agent, it sees only intuitive to implement ReflexVaccumAgent using SimpleReflexAgentProgram backend. In this approach, test is added for SimpleReflexAgentProgram as testing code is already present for ReflexVaccumAgent.)

@norvig and @MrDupin Please review this PR!

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -141,8 +141,8 @@ class Rule:
     """This defines a tempelate for a rule. Different kinds of rules can
     be defined using this template by specifying two attributes:
             i.  'state'  : The current abstracted state of the environment.
-                           This is determined using the agent percepts
-            ii. 'action' : The action to be taken by the agent at the given sate
+                           This is determined using the agent percepts.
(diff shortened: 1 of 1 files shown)


norvig:
I think it is clearer the way it was, rather than adding the `Rule` class. I don't like the `interpret_input` function, because it interacts with the rules: it determines what is important, based on what the rules expect. Any functionality like that should be independent of the rules. Let's just leave things the way they are.

PR #755, opened 2018-02-18
Added virtualenv instructions

### LINKED ISSUE #754: Add virtualenv installation details in CONTRIBUTING.md
If someone is having both python 2.x and python 3.x, it becomes essential to create virtualenv to separate both.
Also new users who are not aware of this may feel stucked. 

This will help users to create virtual environment before getting started.

I will be adding a PR. See if it makes things easier for others. 

Any admin here to comment on this issue?

umangahuja1:
Fixes #754 

Added virtualenv installation details that will ease out testing process too.

@MrDupin have a look

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,6 +101,18 @@ Patch Rules
 Running the Test-Suite
 =====================
 
+Install virtualenv to keep python 2.x and python 3.x separated
+
+    sudo pip3 install virtualenv 
+
+Then create a virtual environment
+
+    virtualenv -p python3 env
+    
+Activate this virtual environment
+
+    source env/bin/activate
+
 The minimal requirement for running the testsuite is ``py.test``.  You can
 install it with:
 
@@ -119,6 +131,10 @@ Fetch the aima-data submodule:
 Then you can run the testsuite from the `aima-python` or `tests` directory with:
 
     py.test
+    
+If you want to deactivate the virtualenv you created, simply use command:
+    
+    deactivate
 
 # Choice of Programming Languages
 


MrDupin:
I am not that familiar with `virtualenv` so I can't provide any concrete feedback.

umangahuja1:
Actually this will help to distinguish modules between python2 and 3 version to avoid any complexities and errors. Anyone else here whom I can refer to validate.

Also I would be thankful if you can tell how should I approach in this community. I am a GSoC aspirant and went through all the guidelines.

apb7:
Hello, @MrDupin! I am quite familiar with  `virtualenv`.   
@umangahuja1: There are a few points of concern (which I feel):  
1. Using `virtualenv` just for testing purposes does not seem to be justified since there is only one module, `pytest` which needs to be installed. This can easily be done by `pip3 install pytest`.  
2. Setting up `virtualenv` takes some time which again could be avoided by using `pip3` instead of `pip`.  3. If we intend to use it for the complete repository, we will have to install all the third-party libraries we use. This might lead to a bit complicated procedure since all we want is that the reader can quickly run the notebooks and go through them.  

Please feel free to correct me at any point. I'll be happy to discuss it :)

umangahuja1:
Most of the part you are correct. 

But I was working with ubuntu and had pytest alreay installed both python2 and python3 version. Hence when I was running it by default as py.test, it was running with python2 and that resulted in 14 errors and that's when I felt this can be case with others too.

We can discuss more since I feel it can happen with others too plus some beginners will also get to learn more about virtualenv this way(maybe).

apb7:
@umangahuja1: I think your problem could be solved by using `pip3 install pytest` instead of `pip install pytest` (which we can change in `CONTRIBUTING.md`).  
Thanks!

umangahuja1:
No I did that part and still it was not solved reason being I had both pytest with python2 and 3 version installed. Hence it was not possible to distinguish between them.

apb7:
@umangahuja1: Okay! Understood :)

umangahuja1:
So I think this will be helpful and also if we at any point of time want to add more dependencies, using virtualenv will be a good practice.

What's your say.

apb7:
@umangahuja1: I think we should leave this for the collaborators to decide for now :)

umangahuja1:
Sure :) At least for me the problem is solved so I will look to contribute more on other issues and pseudocodes.

norvig:
I don't see a compelling argument for why we need to add the complication of `virtualenv`. Everything we do should be in Python 3.

PR #759, opened 2018-02-20
search.ipynb: Remove erroneous output and add RBFS visualisation. 

roberthoenig:
The RBFS visualisation currently uses show_map to display the graph iterations, like the other algorithms. show_map is limited to setting node colors, so we cannot use it (in its current form) to add  f-cost labels to nodes. If it seems reasonable to you to extend the functinality of show_map, let me know, and I'll be happy to do it.


norvig:
The right way to do it is that `def recursive_best_first_search` should not mention anything at all about colors or display. It should just be the algorithm. All the code for display should go into a wrapper around problems. Then you call something like `recursive_best_first_search(DisplayProblem(problem))`.

norvig:
Sorry, it was my fault for delaying in a review, but now there are conflicts. I think we can incrementally do the changes as you suggest. If it is not enough to wrap the problem, we can also make use of `utils.injection`, something like this:

     def DisplayPriorityQueue(...):
         "A PriorityQueue, but handles display of nodes."
         ...

     with injection(PriorityQueue=DisplayPriorityQueue):
         ... search(...) ...

dmeoli:
Thanks @roberthoenig. search.ipynb has changed substantially since 2018; this RBFS-visualization notebook change is stale against the current notebook and hard to apply cleanly. A fresh, focused notebook PR would be welcome. Closing — thanks!

PR #764, opened 2018-02-23
NLP Apps: Author Recognition

MrDupin:
I wrote a new section on `nlp_apps`, which adds a tutorial on recognizing authors using word n-grams. I believe this is a nice addition to the char n-gram tutorial already on the notebook.

### DIFF
--- a/nlp_apps.ipynb
+++ b/nlp_apps.ipynb
@@ -15,7 +15,8 @@
    "source": [
     "## CONTENTS\n",
     "\n",
-    "* Language Recognition"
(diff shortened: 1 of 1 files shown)


norvig:
This is very nice. I like the author ID. Another data set to apply this to is the [Federalist Papers](http://www.gutenberg.org/cache/epub/18/pg18.txt) which are interesting because there are multiple documents on a similar topic area by different authors, and because some of them have disputed authors.

MrDupin:
I will get started on it, sounds very interesting.

PR #765, opened 2018-02-24
Added utils.ipynb

SkAdilina:
utils.ipynb now contains Queue, FIFOQueue and PriorityQueue

### DIFF
--- a/utils.ipynb
+++ b/utils.ipynb
@@ -1,0 +1,175 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": true
+   },
+   "source": [
+    "# Utilities\n",
+    "\n",
+    "This notebook provides some utilities widely used by other modules for the book *Artificial Intelligence: A Modern Approach.* This notebook uses implementations from [utils.py](https://github.com/aimacode/aima-python/blob/master/utils.py) module. Let's start by importing everything from utilities module."
+   ]
+  },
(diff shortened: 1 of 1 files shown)


apb7:
@MrDupin and @SkAdilina : Do we actually require a notebook for utilities? We use notebooks to explain and visualize various ideas and concepts of AI whereas utilities _help_ us to do so. They do not actually represent a concept. Please correct me if I am wrong!  
Thanks!

SkAdilina:
@apb7 I am definitely not an expert on this, so I won't be able to correct you. However, [this](https://github.com/aimacode/aima-python/blob/master/README.md#index-of-algorithms) is where I saw that Queue is yet to be included in the Notebook, so I went ahead and added it.

apb7:
Okay, let's wait for @norvig and @MrDupin then :)

MrDupin:
Hmm, you are right that it seems like we want a notebook for utils with the `Queue` in the README. I personally think such a notebook is not necessary, so maybe we want to remove `Queue` from the list of algorithms, or add a 'Not Needed' mark on the 'Notebook' section.

It is up to @norvig though.

norvig:
I don't think we need a notebook for utilities, including Queue. I think we can just import the code.

MrDupin:
I am afraid since Mr. Norvig said we don't need a notebook for utilities, I have to close this.

Thanks for the work though. Keep it up!

PR #743, opened 2018-02-15
Enhanced mdp notebook

ad71:
- Added missing Policy Iteration section to mdp.ipynb
- Added explanation of policy iteration
- Solved `sequential_decision_environment` using `policy_iteration`
- Added LaTeX for simplified Bellman equations
- Fixed errors when showing pseudocode
- Removed ambiguous test (pointed out in #742 )
- Added section for sequential decision environments
- Solved environments with step by step explanation
- Explained features of the classes in mdp.py with examples
- Added tutorial for `grid_mdp` editor in the appendix section
- Fixed formatting errors

PR #771, opened 2018-02-25
Updated README.md

ad71:
Policy iteration has been added to the mdp notebook in #743 
I forgot to update the readme.
Node has also been added to the search notebook in #761


MrDupin:
Nice catch!

Issue #772, opened 2018-02-25
Better way to structure the repo?


bakerwho:
Perhaps it's just me, but it looks pretty hard to navigate through the code while reading through the book. I have to open and read [agents.py](https://github.com/bakerwho/aima-python/blob/master/agents.py) to see that it discusses code from Chapters 1 and 2.

There's a notebook called [intro.pynb](https://github.com/bakerwho/aima-python/blob/master/intro.ipynb)  that gives some information about the reasons for this - 
> A typical module in the repository has the code for a single chapter in the book, but some modules combine several chapters.
But this notebook is not referenced in the README. Besides, it may be nice to have a companion to the book that links, chapter by chapter, section by section, what files or notebooks might be relevant.

Similarly, the [instructions](https://github.com/aimacode/aima-python/blob/master/SUBMODULE.md) to deal with the [aima-data](https://github.com/aimacode/aima-data/tree/c81e8907917c60bfaedccc720c6b8ce07fabb222) submodule - are not mentioned anywhere in the README. The emptiness of the aima-data subfolder after a `git clone` might be confusing for newcomers to git. If they move files into the folder from local downloads, it might also complicate the merging of their changes (not 100% sure about this, but it looks like it might be a problem).

I'd like to open a discussion on this before I make any changes, but what do you think, @norvig, @MrDupin?

MrDupin:
You are right that we do not have a clear installation guide. I have amended that in #773, thanks for pointing it out!

The instructions in SUBMODULE are simply for updating the submodule to the latest version, so that users can download the correct datasets. It is not something that users need to do themselves, it is more of a reminder for admins.

About the structure, I think it's fine as it is. We already have the README that points people to the right direction. It is up to Mr. Norvig, but I personally wouldn't change this.

JaeDukSeo:
I tried to clean it up, but it wasn't the right approach. So I guess we'll have to wait.

bakerwho:
Might it make sense to put the notebooks in a folder called 'notebooks' and the `.py` files in a folder called 'bin' or 'pycode' or 'scripts'? The repo is still a bit messy. Scrolling through it, one finds files like `pytest.ini` and `requirements.txt` in midst of `ipynb` files that explain concepts! Is there a reason such restructuring has not already been done?

If we're not taking any further steps in this direction, I reckon this issue can be closed.

dmeoli:
[...]

PR #773, opened 2018-02-26
Add Installation Guide to README

MrDupin:
@bakerwho pointed out in #772 that we do not have a clear and easy-to-spot installation guide. With this PR I basically move the little installation guide we had in CONTRIBUTING to the README so that it is harder to miss.


norvig:
Maybe we don't need CONTRIBUTING at all?

MrDupin:
@norvig: I think the CONTRIBUTING does serve a purpose. It is a good place for potential contributors to get a gist of the direction they can help contribute in. Maybe it needs a bit of a refocus though. The "Running the Suite" section should probably be removed (since the section now lies in the README). Also, maybe the "Contributing a Patch/Reporting an Issue/Patch" rules sections are a bit redundant (since most people already follow these rules).

I will set aside some time to work on it this weekend.

PR #776, opened 2018-02-26
Fixed typos and added inline LaTeX to mdp.ipynb

bakerwho:
Added inline LaTeX using `$\\pi$` instead of _pi_. Similarly for other symbols like _gamma_, _delta_, _epsilon_. This also allows us to use LaTeX instead of the clunky Unicode `&#960`.

Also fixed some typos and rewrote some explanations.


mdp.ipynb
```diff
@@ -682,40 +682,40 @@
...
-    "This value of delta decreases over time.\n",
-    "We terminate the algorithm if the delta value is less than a threshold value determined by the hyperparameter _epsilon_.\n",
+    "This value of delta decreases as the values of $U_i$ converge.\n",
+    "We terminate the algorithm if the $delta$ value is less than a threshold value determined by the hyperparameter _epsilon_.\n",
     "\n",
     "$$\\delta \\lt \\epsilon \\frac{(1 - \\gamma)}{\\gamma}$$\n",
     "\n",
-    "To summarize, the Bellman update is a _contraction_ by a factor of `gamma` on the space of utility vectors. \n",
-    "Hence, from the properties of contractions in general, it follows that `value_iteration` always converges to a unique solution of the Bellman equations whenever gamma is less than 1.\n",
+    "To summarize, the Bellman update is a _contraction_ by a factor of $gamma$ on the space of utility vectors. \n",
```
ad71: I think this should not be changed as the backticks were used to denote that `gamma` is a variable used in the code.

mdp.ipynb
```diff
@@ -1565,17 +1600,17 @@
     "We cannot have fixed action sequences as the environment is stochastic and we can eventually end up in an undesirable state.\n",
     "Therefore, a solution must specify what the agent shoulddo for _any_ state the agent might reach.\n",
     "<br>\n",
-    "Such a solution is known as a **policy** and is usually denoted by **&#960;**.\n",
+    "Such a solution is known as a **policy** and is usually denoted by $\\pi$.\n",
```
ad71: I didn't know inline LaTeX was a thing and hence used Unicode. Good job on this one!

mdp.ipynb
bakerwho: Makes sense. Changing it.

bakerwho: Hmm, just went through this again and it seems to be a more mathematical/conceptual statement than an implementational one. In the sense, the $\gamma$ here isn't explaining the code, or even the pseudocode.

I think sticking to the symbol might be the better option. Thoughts?

ad71: You may be right. It's been some time since I wrote this and I don't remember what I was thinking. You can go ahead.

PR #770, opened 2018-02-25
Add test for table_driven_agent_program and Random_agent_program

Noumanmufc1:
I have written test for the tableDrivenAgentProgram. @MrDupin and @norvig  Can you people kindly review it?

PR #749, opened 2018-02-16
Add test for TableDrivenAgentProgram.

roberthoenig:
Fixes #748.

PR #777, opened 2018-02-27
Added TableDrivenAgentProgram tests

ayushjain19:
I was going through PR #770  and #749 for adding tests for TableDrivenAgentProgram.
It seems to me that the main purpose of TableDrivenAgentProgram is to append the "percepts" list in the function and take the decisions accordingly.
It would be great if we add tests to check the status of the environment after **every step** when a new percept is encountered.
This would not only ensure that final environment status is correct but also that the "percepts" list is updated in a right manner.
PR #749 also takes this thinking into account. But, it is always great to add a test which was **discussed in the book**. 
I have implemented the same.

Please have a look @delaray , @MrDupin , @norvig


norvig:
Sorry; an earlier PR introduced Conflicts here.

ayushjain19:
Okay! I will try implementing with respect to the recent PRs merged.

ayushjain19:
Can you please give your inputs now @norvig

Issue #760, opened 2018-02-20
Missing POMDP value-iteration


ad71:
mdp.py has no implementation of POMDP value-iteration. It cannot be directly implemented as we need to build a pomdp class. There are subtle differences between an mdp and a pomdp and I would like suggestions on whether to inherit from the mdp class or implement the pomdp class from scratch. @norvig @MrDupin what do you suggest?

MrDupin:
I think building it from scratch is the better method, in my opinion.

ad71:
I thought so too. I'll get to work as soon as my exams finish this week.

PR #781, opened 2018-02-28
Fix MDP class and add POMDP subclass and notebook

bakerwho:
Work in progress. In order to add a `POMDP` class that can derive from the `MDP` class, we require the MDP class to be made more consistent. For instance, it should not accept arbitrary empty arguments, and should have some consistency check.

I fixed this by restricting the usage of the `MDP` class, and by streamlining calls to it from `GridMDP` and `CustomMDP`. Previously, these derived classes were calling `MDP` and then setting the values of the transition matrix and actions (et cetera) of `MDP` within _their_ respective `__init__()`s. I streamlined this to have only one broad call to MDP, setting all variables. Customization, if required, can still be done by delving into the base MDP class. However, I believe it the purpose of `__init__` as a one-stop setup should be respected.

TODO: `MDP` can now be used as a generic skeleton for a `POMDP` class.

Besides, I have added a POMDP notebook with detailed explanations following the book. Implementation of POMDPs and value iteration to follow. I'm trying to think of an intermediate example between the 2-state case discussed in the book, and the grid search used for MDP. Something that can take advantage of computing power and 3D graphs (or other representations), but still be approachable and intuitive. Any suggestions?

@norvig, @MrDupin Any and all comments are appreciated.


bakerwho:
It seems that the tests here are somewhat problematic.

[tests/test_mdp.py](https://github.com/aimacode/aima-python/blob/master/tests/test_mdp.py) contains `test_transition_model()` which previously listed the transition matrix as a dict keyed by state, of dicts keyed by action, of **tuples** like `(0.4, 'A')`. 
But the convention explained in the comments is that the effects of action $a$ on state $s$ should be a **list of tuples** of the above form. I have changed this test.

[tests/test_rl.py](https://github.com/aimacode/aima-python/blob/master/tests/test_rl.py) is failing on `test_PassiveADPAgent()`. I'm finding this a bit hard to debug as `PassiveADPAgent()` isn't explained in the notebooks, and has thus been implemented, tested, but never actually used for demonstration purposes.

I plan to get to the bottom of this over the next few days. Hopefully `MDP` comes out of this cleaner and easier to use.

ad71:
I think you should have checked open issues first. I am already working on this :(
Have a look at #760

bakerwho:
Whoop @ad71, looks that that was about a week ago. No worries, we can combine efforts. :)

ad71:
Sure @bakerwho . But as @MrDupin suggested, we are better off making a new POMDP class instead of inheriting from MDP. Changing the MDP class will break many tests and the notebooks will have to be rewritten. We just need to write a new class. Your pomdp notebook is nice. We can use that, but I'd suggest closing this PR, update your working directory and start afresh with a new PR.

bakerwho:
@ad71 I feel that as the basic structure of `MDP` (a set of states, a set of actions for each state, a transitional probability on states and actions) will be exactly the same as that of POMDP, and it is duplicating of effort, as well as allowing for inconsistency, if we rewrite that basic stuff for the POMDP class.

Besides, delving into the problems with `MDP` has helped me fix some a foundational bug in `rl.py` which is built on it. As you can see above, the class `PassiveADPAgent` which is built on MDP was failing to update the MDP states after making local copies of them. This caused the tests to fail.

Keeping `MDP` cleaner, and avoiding the duplicating of code - will help prevent such issues in the future. Can your advice on #760 be reconsidered, @MrDupin? @norvig, awaiting your comments.

ad71:
@bakerwho you seem to have a better idea of what you're doing. Let's wait for @MrDupin and @norvig .

bakerwho:
@norvig @MrDupin any comments or should I continue working on this?

mdp.py
```diff
@@ -21,20 +21,36 @@ class MDP:
     list of (p, s') pairs. We also keep track of the possible states,
     terminal states, and actions for each state. [page 646]"""
 
-    def __init__(self, init, actlist, terminals, transitions={}, states=None, gamma=.9):
+    def __init__(self, init, actlist, terminals, transitions = {}, reward = None, states=None, gamma=.9):
```
norvig: No spaces around `=` for keyword arguments.

norvig:
I'll have to study more carefully to see what to do with MDP/POMDP in the long run, but for now, these changes look helpful.

Issue #705, opened 2018-01-31
Implementing a stochastic game example(Backgammon) in games.py


AdityaDaflapurkar:
The Expectiminimax algorithm from the chapter 'Adversarial Search'(section : stochastic games) is not implemented in games.py. Also, it is not added to the list of algorithms. Should I work on it's implementation?

ad71:
As far as I know, expectiminimax is not an algorithm. It is a (mathematical) function that "generalizes the minimax value for deterministic games to an expectiminimax value for stochastic games". It is something like a heuristic for stochastic games and (as far as I know) isn't used by the currently implemented algorithms in games.py
Correct me if I'm wrong.

AdityaDaflapurkar:
Yes, you are right @ad71 . It is not an independent algorithm but just a variant of minimax algorithm for stochastic games. Currently there isn't any implementation for a stochastic game in games.py. Thus Expectiminimax is not used anywhere. @MrDupin ,will it be good if I try implementing a stochastic game(Backgameon) in games.py? I will update this issue accordingly.

MrDupin:
Sure, sounds good!

PR #783, opened 2018-02-28
Backgammon implementation

AdityaDaflapurkar:
@norvig @MrDupin,  I am still working on #705. I have implemented expectiminimax and the basic features of backgammon as given [here](http://www.bkgm.com/rules.html). I have a doubt regarding the computation of utility for a state. In the book, it is given that utility values for backgammon range from 0 to +192. Currently I am returning just +1 or -1 depending upon the player which is winning. Is my approach correct or do I need to use some other function for utility? Also please suggest if there are any other changes which I could make to the code at this stage.


norvig:
I think +1 or -1 is ok. The idea of the extended range is to handle doubling and gammoning properly, but that's not crucial.

norvig:
Also, the other games don't have the equivalent of your `Point` class: I think it would be simpler if the board is just a list of either lists (like `['B', 'B']`) or strings (like `'BB'`).

PR #784, opened 2018-02-28
Added test for simpleProblemSolvingAgentProgram

Noumanmufc1:
@MrDupin  I have implemented the abstract class of simpleProblemSolvingAgentProgram and made a simple vacuum cleaner agent. The states are according to the vacuum cleaner location and also whether the locations are clean or dirty. Please review. The states can be visualized in the following picture:
image

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -199,31 +199,36 @@ def GA_GraphColoringInts(edges, fitness):
     population = init_population(8, [0, 1], 4)
 
     return genetic_algorithm(population, fitness)
+
+
 def test_simpleProblemSolvingAgent():
     class vacuumAgent(SimpleProblemSolvingAgentProgram):
-        def update_state(self, state, percept):
-            state = percept
-            return state
+        def update_state(self, percept):
+            return percept
+
         def formulate_goal(self, state):
             goal = [state7, state8]
             return goal  
+
         def formulate_problem(self, state, goal):
             problem = state
-            return problem       
+            return problem   
+    
         def search(self, problem):
-            if (problem == state1):
+            if problem == state1:
                 seq = ["Suck", "Right", "Suck"]
-            elif (problem == state2):
+            elif problem == state2:
                 seq = ["Suck", "Left", "Suck"]
-            elif (problem == state3):
+            elif problem == state3:
                 seq = ["Right", "Suck"]
-            elif (problem == state4):
+            elif problem == state4:
                 seq = ["Suck"]
-            elif (problem == state5):
+            elif problem == state5:
                 seq = ["Suck"]
-            elif (problem == state6):
+            elif problem == state6:
                 seq = ["Left", "Suck"]
             return seq
+
     state1 = [(0, 0), [(0, 0), "Dirty"], [(1, 0), ["Dirty"]]]
     state2 = [(1, 0), [(0, 0), "Dirty"], [(1, 0), ["Dirty"]]]
     state3 = [(0, 0), [(0, 0), "Clean"], [(1, 0), ["Dirty"]]]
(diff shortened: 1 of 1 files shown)


tests/test_search.py
```diff
@@ -199,7 +199,44 @@ def GA_GraphColoringInts(edges, fitness):
     population = init_population(8, [0, 1], 4)
 
     return genetic_algorithm(population, fitness)
-
+def test_simpleProblemSolvingAgent():
```
MrDupin: You need two add two spaces above this to separate the tests.

tests/test_search.py
```diff
@@ -199,7 +199,44 @@ def GA_GraphColoringInts(edges, fitness):
...
+    class vacuumAgent(SimpleProblemSolvingAgentProgram):
+        def update_state(self, state, percept):
+            state = percept
+            return state
+        def formulate_goal(self, state):
```
MrDupin: You should be adding a space to separate functions. You need to do this for all functions in this test.

tests/test_search.py
```diff
@@ -199,7 +199,44 @@ def GA_GraphColoringInts(edges, fitness):
     population = init_population(8, [0, 1], 4)
 
     return genetic_algorithm(population, fitness)
-
+def test_simpleProblemSolvingAgent():
+    class vacuumAgent(SimpleProblemSolvingAgentProgram):
+        def update_state(self, state, percept):
```
MrDupin: I am not sure why you take `state` as a parameter. The parent class defines the function as `update_state(self, percept)`. Can't you just return `percept`?

tests/test_search.py
```diff
@@ -199,7 +199,44 @@ def GA_GraphColoringInts(edges, fitness):
...
+            state = percept
+            return state
+        def formulate_goal(self, state):
+            goal = [state7, state8]
+            return goal  
+        def formulate_problem(self, state, goal):
+            problem = state
+            return problem       
+        def search(self, problem):
+            if (problem == state1):
```
MrDupin: Remove the parentheses, in this repo we simply write `if condition`.

tests/test_search.py
```diff
@@ -199,7 +199,44 @@ def GA_GraphColoringInts(edges, fitness):
...
+                seq = ["Suck", "Left", "Suck"]
+            elif (problem == state3):
+                seq = ["Right", "Suck"]
+            elif (problem == state4):
+                seq = ["Suck"]
+            elif (problem == state5):
+                seq = ["Suck"]
+            elif (problem == state6):
+                seq = ["Left", "Suck"]
+            return seq
```
MrDupin: Add a space below this line to separate it from the rest of the test.

tests/test_search.py
```diff
@@ -199,7 +199,44 @@ def GA_GraphColoringInts(edges, fitness):
...
+    state1 = [(0, 0), [(0, 0), "Dirty"], [(1, 0), ["Dirty"]]]
+    state2 = [(1, 0), [(0, 0), "Dirty"], [(1, 0), ["Dirty"]]]
+    state3 = [(0, 0), [(0, 0), "Clean"], [(1, 0), ["Dirty"]]]
+    state4 = [(1, 0), [(0, 0), "Clean"], [(1, 0), ["Dirty"]]]
+    state5 = [(0, 0), [(0, 0), "Dirty"], [(1, 0), ["Clean"]]]
+    state6 = [(1, 0), [(0, 0), "Dirty"], [(1, 0), ["Clean"]]]
+    state7 = [(0, 0), [(0, 0), "Clean"], [(1, 0), ["Clean"]]]
+    state8 = [(1, 0), [(0, 0), "Clean"], [(1, 0), ["Clean"]]]
+    a = vacuumAgent(state1)
```
MrDupin: Add a space above this to separate it from the rest.

tests/test_search.py
```diff
@@ -199,7 +199,44 @@ def GA_GraphColoringInts(edges, fitness):
...
+    assert a.__call__(state6) == "Left"
```
MrDupin: In Python, `__call__` simply defines what the object will do when called upon. You can do `a(state6)` and it should work.

tests/test_search.py
Noumanmufc1: The parent class does defines it like this but in the __call__ method the function is passed with both of the arguments. I don't think there is a need for that. Should I keep it like this or change the parent class too? @MrDupin

MrDupin: Ah yes, that is true. I would keep `def update_state(self, state, percept)`, but I would just return `percept` to keep things simple.

Noumanmufc1:
I have made the changes you asked for. @MrDupin  Can you review again?

PR #802, opened 2018-03-04
Added to-cnf explanation

ad71:
- Added brief explanation of `to_cnf` algorithm
- Added usage examples
- Updated README.md


MrDupin:
We mainly reserve the Readme sections for algorithms that are explicitly stated in the book (+ some exceptions). Does `to_cnf` have pseudocode in the book? If not, I think it would be best if we didn't add it to the Readme.

ad71:
Oh, you're right, it doesn't have pseudocode in the book. I misunderstood.

PR #803, opened 2018-03-05
Cleaned the repository

JaeDukSeo:
So I was thinking, let's have the repo look like java for clarity. Also, I think it is a good idea to make a link to not only to the file but also the exact line of the where each class exists. 

FYI, still have to make the python scripts work properly, but I bet this is an easy change since it is changing the import location. 
Please let me know.


JaeDukSeo:
Oh my bad, have to test the code

JaeDukSeo:
Okay, I fixed all of the test cases, now they are all working correctly. Also, the link has to be all configured. Will do that soon.

JaeDukSeo:
The ipython view test case is fixed as well.

norvig:
**No.** We definitely won't be doing this.

`import this` => "Flat is better than nested."

`sys.path.append` is not acceptable.

"look like java" is  not a design goal.

JaeDukSeo:
Gotcha, I see, thanks for letting me know!

PR #733, opened 2018-02-13
Fix EightPuzzle class implementation in search.py (#710)

SasankYadati:
In `search.py`, Environments / Problems are defined as classes which are derived from the `Problem` class. In these changes I made, I rewrote most of `EightPuzzle` class as a derived class of `Problem`. Ran the tests using py.test and also made sure the code is Python 3.4 complaint. Please review my changes and let me know your thoughts. Thanks.

PR #807, opened 2018-03-06
Refactored EightPuzzle class

### LINKED ISSUE #710: EightPuzzle code is not consistent with rest of search.py
The various search algorithms seem to want problems to be defined as subclasses of Problem,
but EightPuzzle is some kind of odd stand-alone section with its own solve method.

It also immediately follows a comment "A* heuristics", but there are no heuristics specified.

### LINKED ISSUE #785: 8-Puzzle problem in search.ipynb throws errors
@MrDupin I reran the search notebook. EightPuzzle doesn't seem to work. I presume this is because of PR #733 .
The PR had quite a few errors which didn't get caught by pytest because tests for EightPuzzle were never written. I plan to fix the errors. Any help would be appreciated.

### LINKED ISSUE #824: Syntax conflicts in search.py (tuples are used as functions!)
In the function `actions()` inside search.py, the variable `index_blank_square` is a `tuple` of indices, but called as a function like `index_blank_square(0)` or `index_blank_square(1)`.
```python
def actions(self, state):
        """Return the actions that can be executed in the given state.
        The result would be a list, since there are only four possible actions
        in any given state of the environment."""
       
        possible_actions = list()
        index_blank_square = self.find_blank_square(state)

        if index_blank_square(0) == 0:
            possible_actions += ['DOWN']
        elif index_blank_square(0) == 1:
            possible_actions += ['UP', 'DOWN']
        elif index_blank_square(0) == 2:
            possible_actions += ['UP']
        
        if index_blank_square(1) == 0:
            possible_actions += ['RIGHT']
        elif index_blank_square(1) == 1:
            possible_actions += ['LEFT', 'RIGHT']
        elif index_blank_square(1) == 2:
            possible_actions += ['LEFT']

        return possible_actions
```

### LINKED ISSUE #826: Another Bug in search.py (iterating over an `int` !)
the function `find_blank_square()` is iterating over an `int`.
In line 422. Here's the code:
```python
def find_blank_square(self, state):
        """Return the index of the blank square in a given state"""
        for row in len(state):
            for column in len(row):
                if state[row][column] == 0:
                    index_blank_square = (row, column)
        return index_blank_square
```
`len(state)` and `len(row)` will _always_ return an `int`.
I wonder how it passed the tests.

Edit: Found two more such cases!
in line 477
```python
    def goal_test(self, state):
        """Given a state, return True if state is a goal state or False, otherwise"""
        for row in len(state):
            for column in len(row):
                if state[row][col] != self.goal[row][column]:
                    return False
        return True
```
and in line 496
```python
    def h(self, state):
        """Return the heuristic value for a given state. Heuristic function used is 
        h(n) = number of misplaced tiles."""
        num_misplaced_tiles = 0
        for row in len(state):
            for column in len(row):
                if state[row][col] != self.goal[row][column]:
                    num_misplaced_tiles += 1
        return num_misplaced_tiles
```

ad71:
- Fixed EightPuzzle class and errors from #733 . It now works with `astar_search` and similar search algorithms.
- Added tests for methods of EightPuzzle
- Added tests for `astar_search` using EightPuzzle
- Closes #785 
- Closes #710 
- Closes #824 
- Closes #826

### DIFF
--- a/search.py
+++ b/search.py
@@ -406,25 +406,25 @@ def astar_search(problem, h=None):
 
 class EightPuzzle(Problem):
 
-    ''' The problem of sliding tiles numbered from 1 to 8 on a 3x3 board,
+    """ The problem of sliding tiles numbered from 1 to 8 on a 3x3 board,
     where one of the squares is a blank. A state is represented as a 3x3 list,
-    where element at index i,j represents the tile number (0 if it's an empty square) '''
+    where element at index i,j represents the tile number (0 if it's an empty square) """
  
     def __init__(self, initial, goal=(1, 2, 3, 4, 5, 6, 7, 8, 0)):
-        ''' Define goal state and initialize a problem '''
+        """ Define goal state and initialize a problem """
 
         self.goal = goal
         Problem.__init__(self, initial, goal)
     
     def find_blank_square(self, state):
-        '''Return the index of the blank square in a given state'''
+        """Return the index of the blank square in a given state"""
 
         return state.index(0)
     
     def actions(self, state):
-        ''' Return the actions that can be executed in the given state.
+        """ Return the actions that can be executed in the given state.
         The result would be a list, since there are only four possible actions
-        in any given state of the environment '''
+        in any given state of the environment """
         
         possible_actions = ['UP', 'DOWN', 'LEFT', 'RIGHT']       
         index_blank_square = self.find_blank_square(state)
@@ -441,8 +441,8 @@ class EightPuzzle(Problem):
         return possible_actions
 
     def result(self, state, action):
-        ''' Given state and action, return a new state that is the result of the action.
-        Action is assumed to be a valid action in the state '''
+        """ Given state and action, return a new state that is the result of the action.
+        Action is assumed to be a valid action in the state """
 
         # blank is the index of the blank square
         blank = self.find_blank_square(state)
@@ -455,12 +455,12 @@ class EightPuzzle(Problem):
         return tuple(new_state)
 
     def goal_test(self, state):
-        ''' Given a state, return True if state is a goal state or False, otherwise '''
+        """ Given a state, return True if state is a goal state or False, otherwise """
 
         return state == self.goal
 
     def check_solvability(self, state):
-        ''' Checks if the given state is solvable '''
+        """ Checks if the given state is solvable """
 
         inversion = 0
         for i in range(len(state)):
@@ -471,8 +471,8 @@ class EightPuzzle(Problem):
         return (inversion % 2 == 0)
     
     def h(self, node):
-        ''' Return the heuristic value for a given state. Default heuristic function used is 
-        h(n) = number of misplaced tiles '''
+        """ Return the heuristic value for a given state. Default heuristic function used is 
+        h(n) = number of misplaced tiles """
 
         return sum(s != g for (s, g) in zip(node.state, self.goal))
 
--- a/search.py
+++ b/search.py
@@ -480,7 +480,7 @@ class EightPuzzle(Problem):
 
         if inversion % 2 != 0:
             check = False
-        print(check)
+        return check
     
     def h(self, node):
         ''' Return the heuristic value for a given state. Default heuristic function used is 
--- a/search.py
+++ b/search.py
@@ -457,9 +457,7 @@ class EightPuzzle(Problem):
     def goal_test(self, state):
         ''' Given a state, return True if state is a goal state or False, otherwise '''
 
-        if state == self.goal:
-            return True
-        return False
+        return state == self.goal
 
     def check_solvability(self, state):
         ''' Checks if the given state is solvable '''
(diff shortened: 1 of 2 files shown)


goswami-rahul:
Great work in the PR!

search.py
```diff
@@ -406,101 +406,92 @@ def astar_search(problem, h=None):
...
-            new_state[blank_square(0)][blank_square(1)] = new_state[blank_square(0)][blank_square(1)+1]
-            new_state[blank_square(0)][blank_square(1)+1] = 0
+        ''' Given state and action, return a new state that is the result of the action.
+        Action is assumed to be a valid action in the state '''
+
+        # ix is the index of the blank square
+        ix = self.find_blank_square(state)
+        new_state = list(state)
+
+        if action == 'UP':
```
norvig: These 8 lines could be

      delta = dict(UP=-3, DOWN=3, LEFT=-1, RIGHT=1)
      neighbor = blank + delta[action]
      new_state[blank], new_state[neighbor] = new_state[neighbor], new_state[blank]

search.py
```diff
@@ -406,101 +406,92 @@ def astar_search(problem, h=None):
...
+            new_state[ix], new_state[ix - 3] = new_state[ix - 3], 0
+        elif action == 'DOWN':
+            new_state[ix], new_state[ix + 3] = new_state[ix + 3], 0
+        elif action == 'LEFT':
+            new_state[ix], new_state[ix - 1] = new_state[ix - 1], 0
+        elif action == 'RIGHT':
+            new_state[ix], new_state[ix + 1] = new_state[ix + 1], 0
         else:
```
norvig: I would say just ignore an illegal action, but if you are going to print something, say what the invalid action is.

search.py
```diff
@@ -406,101 +406,92 @@ def astar_search(problem, h=None):
...
-        for row in len(state):
-            for column in len(row):
-                if state[row][col] != self.goal[row][column]:
-                    return False
-        return True
-
-    def checkSolvability(self, state):
+        ''' Given a state, return True if state is a goal state or False, otherwise '''
+
+        if tuple(state) == self.goal:
```
norvig: Don't do `tuple(state)`, just `state` -- all states must be tuples.

search.py
```diff
@@ -406,101 +406,92 @@ def astar_search(problem, h=None):
...
+
         inversion = 0
         for i in range(len(state)):
-               for j in range(i, len(state)):
-                    if (state[i] > state[j] and state[j] != 0):
-                                    inversion += 1
+            for j in range(i, len(state)):
+                if (state[i] > state[j] and state[j] != 0):
+                    inversion += 1
         check = True
```
norvig: These four lines with `check` should be just
     return (inversion % 2 == 0)

search.py
```diff
@@ -406,101 +406,92 @@ def astar_search(problem, h=None):
...
+            check = False
+        return check
     
-    def h(self, state):
-        """Return the heuristic value for a given state. Heuristic function used is 
-        h(n) = number of misplaced tiles."""
+    def h(self, node):
+        ''' Return the heuristic value for a given state. Default heuristic function used is 
+        h(n) = number of misplaced tiles '''
+
```
norvig: I would do these 5 lines plus 3 blank lines as 1 line:

      return sum(s != g for (s, g) in zip(node.state, self.goal))

tests/test_search.py
```diff
@@ -5,6 +5,8 @@
 romania_problem = GraphProblem('Arad', 'Bucharest', romania_map)
 vacumm_world = GraphProblemStochastic('State_1', ['State_7', 'State_8'], vacumm_world)
 LRTA_problem = OnlineSearchProblem('State_3', 'State_5', one_dim_state_space)
+eight_puzzle = EightPuzzle([1, 2, 3, 4, 5, 7, 8, 6, 0])
```
norvig: In all these tests, states should be tuples, not lists.

ad71:
Thanks for the review @norvig . I'll fix these problems asap.

bakerwho:
@ad71 don't forget to clear outputs from `search.ipynb` to make merging with #812 easier!

search.py
ad71: This is clever. I don't know why I didn't think of this.

ad71:
@bakerwho this PR doesn't alter search.ipynb. Merging #812 should be fine.

ad71:
@norvig The issues you pointed out have been fixed. Thanks!

search.py
```diff
@@ -406,102 +406,75 @@ def astar_search(problem, h=None):
 
 class EightPuzzle(Problem):
 
-    """The problem of sliding tiles numbered from 1 to 8 on a 3x3 board,
+    ''' The problem of sliding tiles numbered from 1 to 8 on a 3x3 board,
```
norvig: No. Pep 257: "For consistency, always use """triple double quotes""" around docstrings."

ad71:
@norvig the docstrings have been fixed

norvig:
Looks good, thanks!

PR #810, opened 2018-03-06
Chapter wise notebooks

bakerwho:
Added a table to [README.md](https://github.com/aimacode/aima-python/blob/master/README.md) with the notebook relevant to each chapter.


MrDupin:
Maybe this is needed, maybe it isn't, I don't know.

Regardless, I think that First Order Logic is covered in `logic.ipynb` and Learning Probabilistic models is in `learning` (where we can find Naive Bayes). Also, Making Simple Decisions has a blank column. For the sake of consistency, I would add a 'None' tag.

Also, in the intro to the section you have three closing parentheses instead of one.

ad71:
I think this isn't needed because the index of the existing table has the chapter number and the names of the notebooks are the same as the names of the corresponding .py files.

bakerwho:
I added this because it felt like a better companion to someone using the Github materials while reading the textbook.

A quick perusal of the website mentioned in the book [aima.cs.berkeley.edu](http://aima.cs.berkeley.edu/) reveals the following:
- The main page links correctly to the Github repo
- This [link](http://aima.cs.berkeley.edu/code.html) has links to the Github repos for 'Overall' and 'Java'. However, the 'Python' link leads to the page it is on, **not** the Github repo.
- The [README](https://code.google.com/archive/p/aima-python/wikis/ReadMe.wiki) linked for Python is outdated.
- The [Google archive](https://code.google.com/archive/p/aima-python/) raises `404 not found` for almost any file. 

This implies that Github is indeed the access point for students who would like to use the code. I thought this might organize the ideas better, seeing how some of the notebooks actually offer nice explanations, intuitions and simulations to accompany the code.

All in all, I feel that a more accessible companion to the book may be a good idea. This was my attempt towards that.

Comments, @MrDupin, @norvig?

norvig:
Let's link to the notebooks from the README, but let's do it from the existing table that now says "Included" for notebooks; let's make that column be the links, not just text.

bakerwho:
Makes sense, on it.

PR #812, opened 2018-03-07
Move viz code + changes to search

### LINKED ISSUE #811: Search Notebook Visualizations
The search notebook currently has a lot of visualizations of graphs and such that are very important to get the concepts across. However, the visualization code itself is not important and the student does not need to read it. For such code we have created the Python script `notebook.py`, which holds the visualization code that goes in the notebooks.

It would be fantastic if we could start looking at ways to move the viz code to `notebook.py`, so that we could clean up the Search notebook a bit. Then, we could simply call the functions from the notebook without showing how the sausage is made.

If you would like to help on this issue, comment below. Thanks!

---

TODO:

* Move `tree_search` and `graph_search` to `notebook.py`. They are mostly visualization code that students don't need to read. We should also have a separate section explaining what the two techniques are, without the visualization. Just walking the student through the code like normal.

bakerwho:
- Moved visualisation methods to notebook.py (fixes #811).
- Made changes for networkx to implicitly read a dict rather than adding node-by-node, edge-by-edge.
- Added outlines to the unexplored white-coloured nodes in `show_map`.
- `search.py` had some bad practices like using 'dict' as a variablename, etc. Fixed these.
- Fixed some formatting.


bakerwho:
Not sure if Travis needs to import the visualisation libraries (will that slow down testing?) However, I'm added those commits anyway.

MrDupin:
Nice work, I think it's good now.

Can you though take care of something I noticed? In the notebook, whenever `display_visual` is called, the line runs a bit too long. I suggest breaking it up a bit. Also, there are spaces around `=`, which should be removed. I am thinking instead of this:

```
display_visual(romania_graph_data, user_input = False, algorithm = depth_first_tree_search, problem = romania_problem)
```

we should do something like:

```
display_visual(romania_graph_data, user_input=False,
               algorithm=depth_first_tree_search,
               problem=romania_problem)
```

search.ipynb
```diff
@@ -1754,74 +1002,44 @@
...
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
    "source": [
     "all_node_colors = []\n",
     "romania_problem = GraphProblem('Arad', 'Bucharest', romania_map)\n",
-    "display_visual(romania_graph_data, user_input = False, algorithm = astar_search, problem = romania_problem)"
+    "display_visual(romania_graph_data, user_input = False, \n",
```
MrDupin: Remove the spaces around the equal signs ( '=' ).

This issue is present in following cells too.

PR #816, opened 2018-03-07
Added test case for CYK_parse

brooks-lee:
I have added a test case for CYK_parse. Please review.

### DIFF
--- a/planning.py
+++ b/planning.py
@@ -74,7 +74,6 @@ class Action:
     def check_precond(self, kb, args):
         """Checks if the precondition is satisfied in the current state"""
         # check for positive clauses
-        #print (self.precond_pos[0].__eq__(expr('At(actor, loc)')), self.precond_neg)
         for clause in self.precond_pos:
             if self.substitute(clause, args) not in kb.clauses:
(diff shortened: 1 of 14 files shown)


MrDupin:
Apart from those, there are some merge conflicts with `agents.ipynb`. That is probably because you are trying to merge a previous version. Copy-paste the current `agents` notebook from the repository into your fork and it should be fixed.

I also noticed that the grammer `E_Chomsky` also has some alignment issues. Can you take care of that too?

Finally, I think I saw that you didn't separate one of the grammars in `nlp.py` from the rest. Can you take care of that as well?

planning.py
```diff
@@ -74,6 +74,7 @@ def substitute(self, e, args):
     def check_precond(self, kb, args):
         """Checks if the precondition is satisfied in the current state"""
         # check for positive clauses
+        #print (self.precond_pos[0].__eq__(expr('At(actor, loc)')), self.precond_neg)
```
norvig: remove this line

brooks-lee: Done.

PR #822, opened 2018-03-08
added definitions of abstract methods and docstrings in utils.py

goswami-rahul:
- Defined the abstract methods for `Queue` abstract class.
- Fixed a bug in `open_data()` which always incorrectly opened file in 'r' mode.
```python
def open_data(name, mode='r'):
    aima_root = os.path.dirname(__file__)
    aima_file = os.path.join(aima_root, *['aima-data', name])

    return open(aima_file)
```
changed last line to
```python
return open(aima_file, mode=mode)
```
- simplified implementation of function `first()`.
- added some docstrings and some typo fixes.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -223,7 +223,7 @@ def rounder(numbers, d=4):
     if isinstance(numbers, (int, float)):
         return round(numbers, d)
     else:
-        assert isinstance(x, collections.abc.Sequence)
+        assert isinstance(numbers, collections.abc.Sequence)
         constructor = type(numbers)     # Can be list, set, tuple, etc.
         return constructor(rounder(n, d) for n in numbers)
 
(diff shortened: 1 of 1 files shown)


goswami-rahul:
Thanks for the detailed feedback @bakerwho  and for sharing the guide for docstrings.
I will revert the changes in the doc strings. Same with the 'rtype' issue.
Will change `queue` to `q`.

I need suggestions on the implementation of `Queue` class.
I am working on the classes `FIFOQueue` and `PriorityQueue` as its subclasses.
Should I change `Stack()` from `function` to a subclass of `Queue`. All three are subclasses of abstract class `Queue` as also mentioned in its doc string, so I think it is a good idea to implement all three as its subclasses for consistency. The usage of `Stack` will remain same.

Another question, exactly what does the `__getitem__` implemented for `PriorityQueue` trying to do (and return)? 
Apparently, it is just returning the same key which called it, if it is present, else None.

goswami-rahul:
Oh, and couldn't the `extend` method in `expr_handle_infix_ops()` you were talking about. @bakerwho

goswami-rahul:
I am closing this PR for now because it has too many reverse changes in doc strings.
I will open a new PR with `Queue` class and its subclasses fully implemented.

PR #825, opened 2018-03-08
This Fixes #824

### LINKED ISSUE #824: Syntax conflicts in search.py (tuples are used as functions!)

goswami-rahul:
The bug where tuples are indexed by `()` instead of `[]` is resolved by this commit.
Also refactored the code, so now the code is shorter and much easier to understand.

### DIFF
--- a/search.py
+++ b/search.py
@@ -426,7 +426,7 @@ class EightPuzzle(Problem):
                 if state[row][column] == 0:
                     index_blank_square = (row, column)
         return index_blank_square
-     
+    
     def actions(self, state):
         """Return the actions that can be executed in the given state.
         The result would be a list, since there are only four possible actions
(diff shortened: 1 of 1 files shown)


ad71:
#824 has already been fixed in #807 . Again, I would implore you to go through the open PRs regarding involving `search.py` and `search.ipynb`

PR #831, opened 2018-03-09
styling and several bug fixes in learning.py

goswami-rahul:
- In many functions, __mutable__ types i.e. `list` were used as the default arguments. These may lead to many bugs and are not recommended because the function evaluates these _default_ values only once at the time of definition and their value may change. So I replaced all such cases with `None` and gave them the default values in the definition.
_Edit_ - opened an __issue__ #833 for this, as it is present in some other files too.

- In many places, when desired output is of `float` type, "Python 2" convention is used to explicitly perform float divisions. for e.g. `1.0/N`. This is now not required in "Python 3", `float` division is implicit with `/` operator.

- Refactored redundant code at some places. PEP and styling changes to match with the rest code.

Suggestions for any additions or changes are appreciated.

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -206,7 +206,8 @@ class DataSet:
             # Find all the item feature values for item in class t
             features = [[] for i in range(feature_numbers)]
             for item in item_buckets[t]:
-                features = [features[i].append(item[i]) for i in range(feature_numbers)]
+                for i in range(feature_numbers):
+                    features[i].append(item[i])
 
             # Calculate means and deviations fo the class
             for i in range(feature_numbers):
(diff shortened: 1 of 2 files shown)


learning.py
```diff
@@ -1060,7 +1059,7 @@ def cross_validation(learner, size, dataset, k=10, trials=1):
             dataset.examples = examples
         return fold_errT/k, fold_errV/k
 
-
+# TODO: The function cross_validation_wrapper need to be fixed.(The while loop runs forever!)
```
MrDupin: You forgot an 's' on 'need'.

goswami-rahul: Thanks for pointing out.

norvig:
I notice you violate PEP 8 by removing spaces around arithmetic operators in some places. I actually don't mind that, but I'm sure others will want to abide by the PEP.

Issue #834, opened 2018-03-11
Test case for double_tennis_problem


brooks-lee:
I have written a test case for the double tennis problem. However, I am proposing a few changes in the double tennis problem implementation. The changes are:
1) Action `Hit(actor, Ball)` can be defined as `Hit(actor, Ball, loc)`; loc is the location the ball is approaching.
2) Action `Go(actor, to)` can be defined as `Go(actor, to, loc)`; loc is the location the actor is currently at; to is the location the actor wants to go to.
3) The goal of the problem can be defined as `required = [expr('Returned(Ball)'), expr('At(a, LeftNet)'), expr('At(a, RightNet)')] ` instead of `required= [expr('Goal(Returned(Ball))'), expr('At(a, LeftNet)'), expr('At(a, RightNet)')]`.

Below is the code for double_tennis problem and the test case for the same:

    def double_tennis_problem():
        init = [expr('At(A, LeftBaseLine)'),
                expr('At(B, RightNet)'),
                expr('Approaching(Ball, RightBaseLine)'),
                expr('Partner(A, B)'),
                expr('Partner(B, A)')]

        def goal_test(kb):
            required = [expr('Returned(Ball)'), expr('At(a, LeftNet)'), expr('At(a, RightNet)')]
            return all(kb.ask(q) is not False for q in required)

        # Actions

        # Hit
        precond_pos = [expr("Approaching(Ball,loc)"), expr("At(actor,loc)")]
        precond_neg = []
        effect_add = [expr("Returned(Ball)")]
        effect_rem = []
        hit = Action(expr("Hit(actor, Ball, loc)"), [precond_pos, precond_neg], [effect_add, effect_rem])

        # Go
        precond_pos = [expr("At(actor, loc)")]
        precond_neg = []
        effect_add = [expr("At(actor, to)")]
        effect_rem = [expr("At(actor, loc)")]
        go = Action(expr("Go(actor, to, loc)"), [precond_pos, precond_neg], [effect_add, effect_rem])

        return PDDL(init, [hit, go], goal_test)

    def test_double_tennis():
        p = double_tennis_problem()
        assert p.goal_test() is False

        solution = [expr("Go(A, RightBaseLine, LeftBaseLine)"),
                    expr("Hit(A, Ball, RightBaseLine)"),
                    expr("Go(A, LeftNet, RightBaseLine)")]

        for action in solution:
            p.act(action)

        assert p.goal_test()

@norvig, @MrDupin Please review.

[...]

PR #837, opened 2018-03-12
test case for double tennis problem

brooks-lee:
I have added a test case for the double tennis problem. The detailed explanation of this can be found in #834. Please review.


norvig:
Shouldn't have things like `#print ` in PRs.

Issue #703, opened 2018-01-29
Add tests for Zebra Problem and Nqueens Problem in test_csp.py


Dragneel7:
Currently there are no tests written for the two algorithms that have been implemented in the file csp.py .
@MrDupin Should we add the tests for these two algorithms.
Thanks.

MrDupin:
Sure, go ahead. Make sure though the tests don't take too long to run.

dubesar:
@antmarakis I think the tests are already added in the test_csp.py file. Please close this issue.

PR #839, opened 2018-03-12
test case for zebra problem

dsaw:
A test case that only checks if the solution of the zebra problem is correct.
_i.e._ Water is drank in the 1st house & Zebra is kept in the 5th house
It's as per issue #703 
Please do review this


norvig:
The build timed out ... maybe the zebra problem takes too much time, I'm not sure. Tests should all be fast.

dsaw:
Hello @norvig ,
It does seem that the random initialization in the build took more time.
Is there any way the tests could run faster? Maybe solving zebra with other algorithms.
Interested to know you thoughts.
Thank you.

Issue #817, opened 2018-03-07
Implement Hybrid Wumpus Agent Algorithm


kunwar31:
Hi, I would like to implement Hybrid Wumpus Agent algorithm, starting with implement the wumpus world environment. 

I'll use this issue to assign myself the task of implementing this algorithm, and complete the code of the environment (if required)

Issue #641, opened 2017-08-31
What I Didn't Do in GSoC 2017


Chipe1:
These are some of the future works I had planned, but couldn't complete during GSoC 2017.
* **Wumpus World:** HybridWumpusAgent needs to be implemented. After this we can display the environment, Wumpus, agent and percepts in the notebook for a better visualization.
* **Eliminate Dependencies:** The base code is mostly independent of external libraries but uses version specific python functions. Such functions (absent or behave differently in older version) need to be identified and fixed. Dependencies like 'networkx' are used in the notebooks and can be replaced by adding code to [notebook.py](https://github.com/aimacode/aima-python/blob/master/notebook.py) which replicates the required functionality. TravisCI testing can also be sped up by excluding unnecessary modules like IPython from travis setup.
* **Expression Parsing:** Currently `Expr` supports having variable number of arguments as a list but functions like `unify()` do not consider nor `expr` make use of this fact.
* **Deploy Notebooks:**  Run the jupyter notebooks on platforms like binder or docker so that it is accessible online to everyone.
* **Notebook Examples**: Many algorithms are yet to be added to the notebooks. The notebooks should include a description of what it does and a few examples of the algorithm being used. Fitting visualization also help a lot.

norvig:
Good projects for someone to work on!

VishnuDuttSharma:
Can you please list down the algorithms for which notebook examples are required to be added?

Chipe1:
@VishnuDuttSharma Few that come to mind are Monte-Carlo-Localization, Policy-Iteration and recently added FOIL, Random Forest.
@norvig Could we add a(nother) column to the table in README.md linking to the section in the notebook where the algorithm is discussed?

norvig:
That's a good idea to have another column in README for notebook.

lenixlobo:
I Would like to contribute to the Random Forest or the Monte-Carlo Localization algorithm.
@Chipe1 @norvig

Chipe1:
@lenixlobo Great. I would suggest you start with random forest. It has the most scope for improvement.

Bala1718:
@Chipe1 I would like to work on Random forest.

lenixlobo:
@Bala1718 I too am currently working on a 
Random Forest notebook.

Bala1718:
@lenixlobo Can we work together?

lenixlobo:
Fine by me @Bala1718

heisenbuug:
[...]

PR #842, opened 2018-03-13
Implemented HybridWumpusAgent

kunwar31:
Solves part of #641 ( implement HybridWumpusAgent ) and therefore part of #817

### DIFF
--- a/logic.py
+++ b/logic.py
@@ -690,6 +690,186 @@ def WalkSAT(clauses, p=0.5, max_flips=10000):
 # ______________________________________________________________________________
 
 
+class WumpusKB(PropKB):
+    """
+    Create a Knowledge Base that contains the atemporal "Wumpus physics" and temporal rules with time zero.
+    """
+    def __init__(self,dimrow):
+        super().__init__()
+        self.dimrow = dimrow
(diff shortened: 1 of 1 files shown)


ad71:
Nice work! Are you going to complete the HybridWumpusAgent class too?

kunwar31:
Yes, I'm going to do that soon, within 1-2 days maximum.

ad71:
Great!

kunwar31:
Thanks @ad71

kunwar31:
@ad71 I've implemented HybridWumpusAgent class, left with plan_route() and plan_shot() helper functions

ad71:
Nice work! After you complete the remaining functions, make sure you add a few simple tests to see if everything works as intended.

norvig:
Looks good. I think a lot more could be done with `Expr` objects rather than string concatenation. That is, rather than `"FacingEast(" + str(t) + ")"` you could just have  `FacingEast(t)` if you have already defined `FacingEast` as an `Expr` function.

kunwar31:
@norvig Thanks for the suggestion. I will try to refactor it.

PR #853, opened 2018-03-15
Added function and test cases for cross-entropy loss

krishnaw14:
I have added function cross_entropy_loss(X,Y) and test_cross_entropy() for cross-entropy loss that is used as loss function for many binary classification problems.

@norvig if it is alright, should I add more error functions?

### DIFF
--- a/tests/test_learning.py
+++ b/tests/test_learning.py
@@ -18,6 +18,16 @@ def test_euclidean():
     distance = euclidean_distance([0, 0, 0], [0, 0, 0])
     assert distance == 0
 
+def test_cross_entropy():
+    loss=cross_entropy_loss([1,0], [0,9, 0.3])
+    assert round(loss,2)==0.23
+
+    loss=cross_entropy_loss([1,0,0,1], [0.9,0.3,0.5,0.75])
+    assert round(loss,2)==0.36
+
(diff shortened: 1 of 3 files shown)


tests/test_learning.py
```diff
@@ -18,6 +18,16 @@ def test_euclidean():
...
+def test_cross_entropy():
+    loss=cross_entropy_loss([1,0], [0.9, 0.3])
+    assert round(loss,2)==0.23
+
+    loss=cross_entropy_loss([1,0,0,1], [0.9,0.3,0.5,0.75])
+    assert round(loss,2)==0.36
+
+    loss=cross_entropy_loss([1,0,0,1,1,0,1,1], [0.9,0.3,0.5,0.75,0.85,0.14,0.93,0.79])
+    assert round(loss,2)==0.26
+
```
goswami-rahul: Add spaces around the `=` and `==`

ad71:
If this isn't in the AIMA book, I don't think we want it in the repository. It is up to @norvig though.

tests/test_learning.py
krishnaw14: Done That. Thank you.

norvig:
We have KL divergence, so cross-entropy makes sense.

Issue #844, opened 2018-03-14
Add Random Forest to learning.ipynb


CharuChhimpa:
@norvig and @MrDupin, Random Forest Algorithm is not added in learning.ipynb. Can I add that in the python notebook?

MrDupin:
Sure, that would be a great addition! Make sure to add the section close to the Decision Tree One.

CharuChhimpa:
@norvig, Can you please review my pr

CharuChhimpa:
@MrDupin please review my pr

CharuChhimpa:
@norvig and @MrDupin, I also find that the functions which are written for testing learners (eg : err_ratio, train_test_split, grade_learner and cross_validation) are no where explained in the python notebook. Are those needed to be added?

MrDupin:
I will review your PR sometime tomorrow, I am a bit busy at the moment.

As for the evaluating functions, I don't think they need to be explained in the notebooks, their docstrings will do just fine.

ad71:
@CharuChhimpa `cross_validation` isn't usable in its current state, nor are the functions that derive from it  (`learning_curve` for example). It won't be useful adding notebook sections for the functions that don't work. See #735 , #730 , #566 and [this issue](https://github.com/aimacode/aima-pseudocode/issues/55) for details. `train_test_split` just splits a dataset into the training and testing set. I can't imagine what a notebook section for this function might look like. You can go ahead if you have some ideas, but hold back on `cross_validation` for now.

PR #855, opened 2018-03-17
Added Random Forest Learner in learning.ipynb

CharuChhimpa:
Refrencing issue #844 
Also added an image file in images folder


ad71:
Nice work! Are you sure we can use the image without giving credits to the original author?

CharuChhimpa:
@ad71 ,Sorry I forgot to mention the souce. Should I mention it?

ad71:
The image is from kaggle, but I don't know if you are allowed to use it without the original author's permission. It may not be a big deal, but I think you can mention the source.

norvig:
Nice work.

PR #820, opened 2018-03-07
Added simpleReflexAgentProgram to vacuumworld.ipynb

Noumanmufc1:
@MrDupin  and @norvig  Please review?

PR #859, opened 2018-03-17
Added simpleReflexAgent to vacuumworld.ipynb

Noumanmufc1:
Fixes PR #820 . Makes use of simple `if else` statement rather than the `Rule` class . @norvig  Please review?

### DIFF
--- a/vacuum_world.ipynb
+++ b/vacuum_world.ipynb
@@ -73,7 +73,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 38,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


vacuum_world.ipynb
```diff
@@ -341,25 +479,45 @@
...
     "# TODO: Implement these functions for two-dimensional environment\n",
     "# Interpret-input function for the two-state environment\n",
-    "def interpret_input(percept):\n",
-    "    pass\n",
-    "\n",
-    "rules = None\n",
+    "def interpret_input(state):\n",
+    "    return state\n",
     "\n",
     "# Rule-match function for the two-state environment\n",
```
norvig: We don't need any of this rule mechanism.  All we need is:

     def VacuumAgentProgram(percept):
        loc, status = percept
        return ('Suck' if status == 'Dirty' else
                'Right' if loc = loc_A else
                'Left')

Noumanmufc1:
Removed all of the `Rule` mechanism @norvig

PR #862, opened 2018-03-18
Refactored WumpusKB and HybridWumpusAgent to use Expr obejcts

kunwar31:
See #842. Solves part of #641 and #817 

Changes : 

-  added Expr functions
-  Refactored WumpusKB to use Expr functions 
-  Refactored HybridWumpusAgent to use Expr functions
-  Fixed minor typos

What's Left ( to solve #817 ) :

-  plan_route and plan_shot implementation
-  Tests for the agent

### DIFF
--- a/logic.py
+++ b/logic.py
@@ -970,7 +970,6 @@ class WumpusKB(PropKB):
 
     def ask_if_true(self, query):
         return pl_resolution(self, query)
-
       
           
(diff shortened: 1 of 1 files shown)


norvig:
Looks good. Maybe `implies_and_implies` could be renamed something like `equiv`.

PR #873, opened 2018-03-21
Added test case for ModelBasedReflexAgentProgram

jl1914:
add a cautious model for agent to update state and guess the environment.


norvig:
We're not using the Rule class and interpreter, so this test doesn't make sense any more.

PR #887, opened 2018-03-24
NLP Apps: Federalist Papers

MrDupin:
I worked on classifying the disputed Federalist Papers, as suggested [here](https://github.com/aimacode/aima-python/pull/764#issuecomment-368269303). I used a simple approach of Word Models + Naive Bayes. One of the issues I encountered was the fact that a lot of multiplications of small floating points are needed. These multiplications oftentimes ended up as 0 instead of a very small number. I am not sure how to fix this properly, so any feedback is much appreciated.

In this case, I solved this by sampling and sub-sampling the to-be-classified examples until the multiplication returned something other than 0. To stabilize the predictions, for each example I make multiple predictions and average them. This seems to be working fine, but is not perfect.

Apart from that, I did not have much time to analyze the data and mess around with different architectures/word models. I believe this will be an excellent project by GSoC students.

If/when this gets merged, I will open an Issue so that others can expand this. I have left enough room so adding to this section will be easy.

### DIFF
--- a/nlp_apps.ipynb
+++ b/nlp_apps.ipynb
@@ -16,7 +16,8 @@
     "## CONTENTS\n",
     "\n",
     "* Language Recognition\n",
-    "* Author Recognition"
+    "* Author Recognition\n",
+    "* The Federalist Papers"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


norvig:
The floating-point-underflow problem is annoying. Floating point only goes down to about 10**-300. With long text passages, that's not enough. The two main options:
- Use `decimal.Decimal` with, say, 3000 digits of precision.
- Add logarithms of probabilities rather than multiplying probabilities.

nlp_apps.ipynb
```diff
@@ -371,6 +372,410 @@
...
+      "Paper No. 64\n",
+      "Hamilton: 0.5756065218890194\n",
+      "Madison : 0.3648418106830272\n",
+      "Jay     : 0.059551667427953384\n",
+      "----------------------\n"
+     ]
+    }
+   ],
+   "source": [
+    "for d in disputed:\n",
```
norvig: I would probably do this something like

     for d in disputed:
         results = ['{}: {:.2f}'.format(name, probs[(name, 1)]) for name in 'Hamilton Madison Jay'.split()]
         print('Paper No. {}: {}'.format(d, ' '.join(results)

PR #888, opened 2018-03-25
NLP Apps: Federalist Decimals

MrDupin:
After feedback on #887 I added the `decimal` library to help do the heavy lifting for probability multiplication. That library is much more powerful than I thought, thanks for letting me know!

Also, I changed the way the results were printed.

Maybe when I get the time I will take a look at logarithms of probabilities, in case it is more elegant than the current way.


norvig:
This looks good. I note that now all the results have a probability of 1.0 for one author, and 0.0 for the others. This is actually not surprising. With many words, the probabilities of the whole document are low for any model, so if it is 10**-500 for Madison and 10**-503 for Jay, that doesn't seem too different, but to two decimal places it rounds off to 1.0 and 0.0. We could add more decimal places, or we could report log_10 probabilities: then Madison would be 0 and Jay would be -3.

MrDupin:
I agree that using logarithms will probably work better. As a matter of fact, the final probabilities round up to 1.0 for up to 10-15 decimal points, using `decimal`.

I will raise an issue so that students can work on this. I think there is a lot of potential here.

Issue #898, opened 2018-03-29
Empty action list for a non-terminal state not handled in player functions 


AdityaDaflapurkar:
@MrDupin, @norvig The player functions in games.py currently throw an error and the execution terminates if at a certain non-terminal game state, the player has no legal moves i.e. the action list is empty. Ideally a player should lose his turn and the game should continue if there are no legal moves. How can this situation be handled?

PR #904, opened 2018-04-01
Add play_game function and fix backgammon game play issues

AdityaDaflapurkar:
@norvig, While testing the backgammon game play between expectiminimax player and query player, I was facing the issue #898. So I have added checks for empty action lists in the player functions. Also currently I have written a play_game() function in backgammon class for testing the game play. Please review and suggest if there are any changes to be made.


games.py
```diff
@@ -41,6 +41,9 @@ def min_value(state):
 
 # ______________________________________________________________________________
 
+dice_rolls = list(itertools.combinations_with_replacement([1, 2, 3, 4, 5, 6], 2))
+direction = {'W' : -1, 'B' : 1}
+
 def expectiminimax(state, game):
```
norvig: I know this is not your fault, somehow the code got like this before, but `expectiminimax` **cannot** be limited to games with two dice. It should be a line-for-line implementation of the pseudocode:

      EXPECTIMINIMAX(s) =
         UTILITY(s)                        if TERMINAL-TEST(s)
        max_a EXPECTIMINIMAX(RESULT(s, a)) if PLAYER(s)= MAX
        min_a EXPECTIMINIMAX(RESULT(s, a)) if PLAYER(s)= MIN
       ∑_r P(r)EXPECTIMINIMAX(RESULT(s, r)) if PLAYER(s)= CHANCE

games.py
```diff
@@ -181,18 +182,21 @@ def query_player(game, state):
     game.display(state)
     print("available moves: {}".format(game.actions(state)))
```
norvig: Alphabeta_search **cannot** have print statements in it, and cannot query the user. Rather, there can be a player who is a user; that player does the `input('Your move? ')` and does the `print`s.

AdityaDaflapurkar: The print statements are currently inside query_player.

Issue #903, opened 2018-04-01
add have_cake_and_eat_cake_too in planning notebook


nvinayvarma189:
I see that there is no example for `have_cake_and_eat_cake_too` in planning notebook. I would like to add it.

MrDupin:
Go ahead. :)

PR #906, opened 2018-04-02
added have_cake_and_eat_cake_too

nvinayvarma189:
This refers to #903. The example for `have_cake_and_eat_cake_too` is added. Also, In the description under `%psource Action`, I have changed `effect_neg` to `effect_rem` to keep it consistent with the rest of the examples in the notebook. Can you please review this @norvig and @MrDupin. Please let me know if there are any changes required. Thanks.


planning.ipynb
```diff
@@ -594,6 +594,104 @@
...
+   "outputs": [],
+   "source": [
+    "have_cake_and_eat_cake_too = have_cake_and_eat_cake_too()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First let us check wether the goal state (have cake and eat cake) is reached or not."
```
MrDupin: It's 'whether', not 'wether'.

planning.ipynb
```diff
@@ -594,6 +594,104 @@
...
+   "metadata": {},
+   "source": [
+    "## Have Cake and Eat Cake Too"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This problem involves the task of eating a cake with an initial condition of having a cake. "
```
MrDupin: This is too short of an introduction. As someone who is not familiar with the example, I am confused as to what the problem statement is. Do we simply have a cake and we want to eat it? Or is there something else to it?

Notebook sections are meant to educate people, not simply show the code. You need to be adding more to flesh your sections out.

This goes for your other sections as well. They are just too short. When students read this, they will have questions this notebook will not answer. More detail is needed, sorry.

nvinayvarma189: Thank you for your suggestion @MrDupin. I have written this example assuming that the reader has the book, as suggested by @norvig. Also, In the book, this `have_cake_and_eat_cake_too` problem is not explicitly described as the other planning problems like the `air_cargo` or the `spare_tire` problem. This problem is actually used as an example for explaining planning graphs.

Initially, I thought of explaining the planning graph with the help of `have_cake_and_eat_cake_too` in the notebook, but then, that explanation would need a bit lengthy description and I assumed that it would not be desirable to put it in the notebook. If it is advisable, then I would add the planning graph explanation and use the `have_cake_and_eat_cake_too` to describe it.

I've taken reference from the other problems like `air_cargo`, `spare_tire` to see how they are implemented and I've written this accordingly. I understand the need to give a detailed description and I will try to provide some more information. Sorry for the inconvenience. Please let me know what you think @MrDupin.

planning.ipynb
nvinayvarma189: Thanks for pointing it out.

planning.ipynb
MrDupin: Even though the reader will probably have the book, I find that it is still useful adding complete information about the problem. There is no need to write paragraphs on end, but clearly and cleanly defining the problem is key in making the section useful. Students shouldn't have to go back and forth between the notebook and the book while reading a section just to understand what the problem is about, that is inconvenient. If the students have the minimum needed information on the notebook, they can use it as a supplement to their studying.

I noticed that the Air Cargo and Spare Tire problems also have the same issue. I will shortly open an issue asking for help on those. The Three Block Tower problem on the other hand is a much better gauge of what we are looking for. A short paragraph explaining what is going on will go a long way in helping students.

nvinayvarma189: Thanks @MrDupin. I will work on it. Also, for the `have_cake_and_eat_cake_too` problem, I will try to use it as an example for explaining planning graphs. I think it actually needs a bit longer description than what is ideal. I will try to shorten it without compromising on the understanding of the reader.

nvinayvarma189:
@norvig and @MrDupin Now I have added more information explaining the problems. This refers to #907. I observed that in `planning.py` there is a separate class which supports planning graphs and I think it will be paired best with the `Graphlan` algorithm. So I haven't included it now. 

Towards the end of `planning.py` there is a test case for `three_block_tower` written. I have removed it.

There is a small issue with `psource` that bothers me. When I run a `psource(function)` cell, It gives me the source code of the function, but when I run it again. It is throwing a lot of errors. I did not find the reason. Can you please check it. Thanks.

planning.ipynb
```diff
@@ -302,13 +303,185 @@
...
+    "\n",
+    "**Cargo(x):** Declare **' x '** as cargo.\n",
+    "\n",
+    "**Plane(x):** Declare **' x '** as plane.\n",
+    "\n",
+    "**Airport(x):** Declare **' x '** as airport.\n",
+    "\n",
+    "\n",
+    "\n",
+    "we have an `initial_state` of having cargo C1, plane P1 at airport SFO and cargo C2, plane P2 at airport JFK. Our goal state is to have cargo C1 at airport JFK and cargo C2 at airport SFO. We will discuss on how to achieve this. Let us now define an object of `air_cargo` problem:"
```
MrDupin: The first sentence is a bit awkward, there are two 'have's in there. I suggest changing it to something like this:

"In the `initial_state` we have cargo C1, plane P1 ..."

I think this reads better.

planning.ipynb
```diff
@@ -342,22 +515,29 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As we can see, it hasn't completed the goal. Now, we define the sequence of actions that it should take in order to achieve\n",
-    "the goal. Then the `airCargo` acts on each of them."
+    "It will give us False because the goal state is not yet reached. Now, we define the sequence of actions that it should take in order to achieve the goal. Then the `airCargo` acts on each of them.\n",
```
MrDupin: "It returns `False`" is better than "It will give us False", in my opinion.

planning.ipynb
```diff
@@ -342,22 +515,29 @@
...
+    "\n",
+    "Let us have an idea of the actions available to us i.e, Load, Unload, Fly\n",
```
MrDupin: Instead of this sentence, write something like: "The actions available to us are the following:"

planning.ipynb
```diff
@@ -342,22 +515,29 @@
...
+    "\n",
+    "**Load(c, p, a):** Load cargo **' c '** into plane **' p '** from airport **' a '**.\n",
```
MrDupin: For this line and the two lines below, you have single quotation marks around the variable names, but you leave a space around the variable name. This can be removed, I think it will look better.

planning.ipynb
```diff
@@ -372,22 +552,19 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "True\n"
+     ]
     }
    ],
    "source": [
-    "airCargo.goal_test()"
+    "print (airCargo.goal_test())"
```
MrDupin: Remove the space after `print`.

planning.ipynb
```diff
@@ -606,12 +1116,153 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This problem involves the task of eating a cake with an initial condition of having a cake. "
+    "This problem involves the task of eating a cake with an initial condition of having a cake. First let us look at take a look at `have_cake_and_eat_cake_too`"
```
MrDupin: You accidentally added the same sentence twice. It should be "First let's take a look at `have_cake...`"

planning.ipynb
```diff
@@ -408,12 +585,169 @@
...
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "psource (spare_tire)"
```
MrDupin: Remove the space after `psource`.

planning.ipynb
```diff
@@ -507,12 +848,175 @@
...
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "psource (three_block_tower)"
```
MrDupin: Remove the space after `psource`.

planning.ipynb
```diff
@@ -606,12 +1116,153 @@
...
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "psource (have_cake_and_eat_cake_too)"
```
MrDupin: Remove the space after `psource`.

MrDupin:
I am not sure what is happening with `psource`, it runs fine on me. What is the error?

nvinayvarma189:
@MrDupin  I've fixed the styling issues. Can you review it now. I'm a bit confused with the `psource` issue. Now it is working fine. I will raise an issue if it occurs again, Sorry.

planning.ipynb
```diff
@@ -594,6 +1104,249 @@
...
+   ],
+   "source": [
+    "psource(have_cake_and_eat_cake_too)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Have(x):** Declraes that we have **' x '**."
```
MrDupin: Noticed this typo: 'Declraes' instead of 'Declares'.

nvinayvarma189: Thanks. Fixed that.

planning.ipynb
```diff
@@ -594,6 +1104,249 @@
...
+   "metadata": {},
+   "source": [
+    "## Have Cake and Eat Cake Too"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This problem involves the task of eating a cake with an initial condition of having a cake. First, let us look take a look at `have_cake_and_eat_cake_too`"
```
MrDupin: The duplicate sentence problem is still present here. Replace the final part ("First, let us ...") with something like "First, let's take a look at ..."

nvinayvarma189: Thanks, I fixed it

planning.ipynb
```diff
@@ -1116,7 +1116,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This problem involves the task of eating a cake with an initial condition of having a cake. First, let us take a look at the `have_cake_and_eat_cake_too`"
+    "This problem involves the task of eating a cake with an initial condition of having a cake. First, let us take a look at `have_cake_and_eat_cake_too`"
```
MrDupin: Looks good now. One final thing, maybe add a colon (':') at the end of the line. Other than that, it looks good.

PR #921, opened 2018-05-08
[WIP] Refactor planning.py

### LINKED ISSUE #903: add have_cake_and_eat_cake_too in planning notebook

### LINKED ISSUE #907: Planning Notebook: Air Cargo and Spare Tire Problems
In the `planning` notebook, the Air Cargo and Spare Tire Problems are not explained very well. The problem statements are not as clear as they could be. They need to be explained in a bit more detail, so that readers can easily understand what is going on without having the book right by their side, going back and forth between book and notebook.

So, I suggest the introduction to these problems is fleshed out a bit more. There is no need for long paragraphs, just a short one like the Three Block Tower Problem will do.

ad71:
Earlier, `GraphPlan` worked _only_ on the `spare_tire` problem. On all the other planning problems, it provided a wrong solution or no solution at all. For example: `air_cargo` gave the solution `[Unload(C1, P1, JFK), Unload(C2, P2, SFO)]`. This solution is incomplete. Furthermore, the graph for `have_cake_and_eat_cake_too` never contained the action `Bake(Cake)`. This usually happened when either preconditions or effects of an action had negative clauses, since `GraphPlan`, `Graph`, `Level` and `Action` classes were built to handle two knowledge bases, one for positive clauses and one for negative clauses. In the `Level` class, these were sometimes mixed up into a single list causing errors in finding mutex links. Nomenclature of `Persistence` actions were problematic too.

This PR solved these problems as follows:
- [x] A new class `UnaryAction` has been created that operates on a single knowledge base and negative preconditions and effects are preceeded by a `Not` in their name, similar to clauses in propositional logic.
- [x] The `Level`, `Graph` and `GraphPlan` classes have been re-written, integrating the positive and negative knowledge bases into a single knowledge base.
- [x] `air_cargo`, `spare_tire`, `three_block_tower`, `have_cake_and_eat_cake_too` have been re-written using `UnaryAction`s and positive effects and preconditions.
- [x] `spare_tire_graphplan` has been re-written with the above changes in mind
- [x] `have_cake_and_eat_cake_too_graphplan`, `three_block_tower_graphplan` and `air_cargo_graphplan` functions solve their respective planning problems using GraphPlan.
- [x] A `refine_solution` function was added to convert the unordered solutions to a linear solution, much like what a total-order planner would output.
- [x] A `shopping_problem` and its solution was added as an example.
- [x] Included tests for `GraphPlan`.
- [x] Modified `test_graph_call` according to the points above.
- [x] Tests have been added for all the newly added functions.
- [x] The planning notebook has been re-written according to these changes.
- [x] A detailed notebook section for `GraphPlan` and its helper classes was added.
- [x] Redefine PDDL
- [x] Combine `Action` and `UnaryAction`
- [x] Incorporate `bound_symbols`

Closes #907 
Closes #903 

`GraphPlan` is fully functional now and gives the correct solution for all problems given in the book. (Though it gives incomplete solutions for complicated problems sometimes, which will be ironed out later)


norvig:
Thanks for this work. A few questions:

- Why use `NotAt(x, y)` -- couldn't you use `~ At(x, y)` ?
- I don't understand why you need a `UnaryAction`. How is this different from `Action`? I realize that there were bugs in the previous implementation that you fixed, but could you have done that by fixing `Action` rather than introducing `UnaryAction`?

ad71:
I understand all your points. I'll take care of these in the next update.

norvig:
Excellent! Thanks.

ad71:
@norvig There is another problem. `FolKB` is used as a knowledge base in all planning problems, but when FolKB encounters a clause that doesn't begin with an alphabet, it raises and exception "Not a definite clause". Hence, `~At` has to be represented as `NotAt` internally. This can be solved in the following ways:
1. Create a new `PlanningKB` class within `planning.py` to handle such clauses. This is a simple solution but would lead to duplication of code and there will be no real use of this class (apart from checking variable naming conventions).
2. Convert and reconvert back and forth between the two representations as and when required. This is a good solution, but requires quite a lot of work in a lot of places, even in the later algorithms.
3. Get rid of the idea of a knowledge base altogether and use a list to store the clauses.

What do you think will be the best solution?

ad71:
Changing the definition of `PDDL` will break the later algorithms in the module, but since those algorithms require a revisit anyway (`hierarchical_search` doesn't work either), it will be better to remove the `UnaryAction` class right now and rebuild everything from the top down and refactor wherever required, as almost nothing works correctly in the planning module. I'm on it.

norvig:
This is nice; good work.  A few things:

- Action.convert does the same thing to two arguments; it would be better to have it do one thing, and call it twice.
- The expression `~At(obj, loc)` is a unary expression whose `op` is `~` and whose first arg is `At(obj, loc)`: `Expr('~', Expr('At', obj, loc))`.  So you can leave things as you have them, or, I think it would be slightly better if you allowed the inputs to PDDL and Action to be either strings or Exprs, and you could do something like this:
```
def convert_literal(literal):
    """If a literal is of the form ~P(X), convert it to NotP(X)."""
    if literal.op == '~':
        arg = literal.args[0]
        return Expr('Not' + arg.op, arg.args[0])
```

ad71:
@norvig I'll incorporate your points into the next update, thanks!

PR #923, opened 2018-05-15
Minor update to planning.py

ad71:
- [x] PDDLs can now be defined using Exprs aswell as strings
- [x] Fixed `goal_test` function
- [x] Added `socks_and_shoes` problem to be used as a simple example for `partial-order-planner` later
- [x] Minor bugfixes

### DIFF
--- a/planning.py
+++ b/planning.py
@@ -23,7 +23,10 @@ class PDDL:
     def convert(self, clauses):
         """Converts strings into exprs"""
         if not isinstance(clauses, Expr):
-            clauses = expr(clauses)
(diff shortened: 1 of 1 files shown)


planning.py
```diff
@@ -207,6 +213,25 @@ def shopping_problem():
                                 effect='At(y) & ~At(x)')])
 
 
+def socks_and_shoes():
```
norvig: This is good, but you could do it without a function:

`socks_and_shoes = PDDL(...)`

PR #929, opened 2018-06-03
Added POMDP-value-iteration

### LINKED ISSUE #760: Missing POMDP value-iteration

ad71:
- [x] Added `POMDP` class
- [x] Added `pomdp_value_iteration`
- [x] Added tests for `pomdp_value_iteration`
- [x] Added `plot_pomdp_utility` helper function
- [x] Merged `pomdp.ipynb` and `mdp.ipynb`
- [x] Added notebook section for `POMDP`
- [x] Added notebook section for `pomdp_value_iteration`
- [x] Added `POMDP` section to `mdp_apps.ipynb`
- [x] Updated README.md
- [x] Removed numpy dependency
- [x] Minor formatting fixes
- [x] Closes #760


mdp.py
```diff
@@ -238,13 +252,206 @@ def policy_iteration(mdp):
...
+            sum1 = 0
+            for element in U1[k]:
+                sum1 += sum(element)
+            sum2 = 0
+            for element in U2[k]:
+                sum2 += sum(element)
+        return abs(sum1 - sum2)
+
+        
+class Matrix:
```
norvig: Not sure ... maybe we should use numpy.matrix ?

MrDupin: If we are to use `numpy` here, we should consider using `numpy` on the rest of the repository as well. It would be an interesting undertaking and I am all for it, but adding another dependency is something that needs discussion.

Maybe an Issue can be opened to discuss this when this is more concrete?

ad71: I thought we aren't supposed to use external libraries, but using `numpy` will make room for a lot of improvements in many algorithms. I think this can be discussed in a new issue as @MrDupin suggested.

PR #933, opened 2018-07-01
Search notebook update

ad71:
- [x] Added tests for `recursive_best_first_search`
- [x] Added tests for `OnlineDFSAgent`
- [x] Added tests for `LRTAStarAgent`
- [x] Minor formatting fixes and bugfixes in `search.py`
- [x] Added visualization for `recursive_best_first_search`
- [x] Added notebook section for `recursive_best_first_search`
- [x] Added notebook section for `and_or_graph_search`
- [x] Added notebook section for `OnlineDFSAgent`
- [x] Added notebook section for `LRTAStarAgent`
- [x] Updated `display_visual` function in `notebook.py`
- [x] Renamed some visualization functions and fixed formatting errors
- [x] Updated README.md


tests/test_search.py
```diff
@@ -151,7 +152,33 @@ def test_conflict():
 def test_recursive_best_first_search():
     assert recursive_best_first_search(
         romania_problem).solution() == ['Sibiu', 'Rimnicu', 'Pitesti', 'Bucharest']
+    assert recursive_best_first_search(
+        EightPuzzle((2, 4, 3, 1, 5, 6, 7, 8, 0))).solution() == [
+            'UP', 'LEFT', 'UP', 'LEFT', 'DOWN', 'RIGHT', 'RIGHT', 'DOWN'
```
norvig: This is probably ok, but are you sure that there isn't another solution of equal length?

ad71: @norvig , there is in fact another solution of equal length
```python
['LEFT', 'UP', 'UP', 'LEFT', 'DOWN', 'RIGHT', 'DOWN', 'RIGHT']
```
but this algorithm (or any search algorithm for that matter) gives one particular optimal solution every time, even if there exists another one of the same length. This happens due to an implementational detail in `EightPuzzle`.
In the `EightPuzzle` class, the `actions` method returns the possible actions, but the actions are stored in a list and every time `node.expand(problem)` is called from RBFS, the neighbors are traversed in the order [up, down, left, right], which is why the algorithm converges to the same solution every time. I also ran it for 1000 cycles just to make sure.
This can be fixed (I don't think it requires fixing) by returning a `set` of possible actions instead of a `list`, but a lot of tests will be rendered useless.

norvig: To keep it simple, I'll accept this as is, but it would be a better test if it didn't rely on an arbitrary choice of ordering.

norvig: That is, the test could be something like : `achieves_goal(start, s) and len(s) == 8`

ad71: Sure, I'll take care of this in the next update

PR #940, opened 2018-07-13
Angelic_search

MariannaSpyrakou:
Added angelic_search in planning.py, tests in test_planning.py  and notebook planning_angelic_search.ipynb,

-- Refinements function supports high level plans with multiple implementations. 
-- Angelic search doesn't work correctly for all instances 
-- Making progress is not implemented yet

### DIFF
--- a/planning.py
+++ b/planning.py
@@ -1253,7 +1253,7 @@ class Problem(PlanningProblem):
             raise Exception("Action '{}' not found".format(action.name))
         self.init = list_action.do_action(self.jobs, self.resources, self.init, args).clauses
 
-    def refinements(hla, state, library):  # TODO - refinements may be (multiple) HLA themselves ...
+    def refinements(hla, state, library):  # refinements may be (multiple) HLA themselves ...
         """
         state is a Problem, containing the current state kb
         library is a dictionary containing details for every possible refinement. eg:
@@ -1292,7 +1292,6 @@ class Problem(PlanningProblem):
         e = Expr(hla.name, hla.args)
         indices = [i for i, x in enumerate(library['HLA']) if expr(x).op == hla.name]
         for i in indices:
-            # TODO multiple refinements
             precond = []
             for p in library['precond'][i]:
                 if p[0] == '~':
@@ -1305,9 +1304,8 @@ class Problem(PlanningProblem):
                     effect.append(expr('Not' + e[1:]))
                 else:
                     effect.append(expr(e))
-            action = HLA(library['steps'][i][0], precond, effect)
-            if action.check_precond(state.init, action.args):
-                yield action
+            action = [HLA(library['steps'][i][j], precond, effect) for j in range(len(library['steps'][i]))]
+            yield [h for h in action if h.check_precond(state.init, h.args)]
 
     def hierarchical_search(problem, hierarchy):
         """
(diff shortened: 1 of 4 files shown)


ad71:
The Travis CI build fails because of some problematic tests in `mdp.py`. These have been fixed in #942 and the build will pass once that is merged.

planning.py
```diff
@@ -1346,6 +1344,160 @@ def result(problem, action):
...
+                if j>0:
+                    prefix = plan.action[:j-1]
+                    outcome = Problem.result(problem, prefix[-1])
+                else: 
+                    prefix = []
+                    outcome = copy.deepcopy(problem)
+                if j < len(plan.action)-1:
+                    suffix = plan.action[j+1:]
+                else: 
+                    suffix = []
```
delaray: Please put the code that computes the prefix and suffix in a separate function. This will allow you to write a unit test for it and it will make the angelic_search function more readable.
i.e. prefix, suffix = find_prefix_and_suffix (....)

planning.py
```diff
@@ -1346,6 +1344,160 @@ def result(problem, action):
...
+                        else: 
+                            reachable_set[i+1].append(state)
+        return reachable_set
+
+
+	
+    def making_progress(plan, initialPlan):
+        """ 
+        Not correct 
+        """
```
delaray: It's good that you noted that this isn't working. Perhaps you can comment on what you think the problem is, or what would be needed to make it correct.

planning.py
```diff
@@ -1435,3 +1587,105 @@ def go_to_sfo():
...
+        super().__init__(action, precond, effect, duration, consume, use)
+
+
+    def convert(self, clauses):
+        """Converts strings into Exprs""" 
+        lib = {'~': 'Not', 
+                '$+': 'PosYes',
+               '$-': 'PosNot',
+               '$$' : 'PosYesNot'}
+
```
delaray: Thank you for making this declarative knowledge.

planning.py
```diff
@@ -1435,3 +1587,105 @@ def go_to_sfo():
...
+                        effects[i] = Expr(clause.op[w:], clause.args[0])
+                        if n==3: effects[i+len(effects)//3] = Expr(clause.op[6:], clause.args[0])
+
+                    else: 
+                        effects[i] = expr(clause.op[w:])
+                        if n==3: effects[i+len(effects)//3] = expr(clause.op[6:])
+            #print('effects',  effects)
+        
+        return [ HLA(Expr(self.name, self.args), self.precond, effects[i] ) for i in range(len(effects)) ]
+
```
delaray: Some comments in the function angelic_action would be really nice. It might be nice to have a separate function that computes and returns n & w. Again this would allow you to test that aspect of the code and improve readability of angelic_action.

planning.py
```diff
@@ -1435,3 +1587,105 @@ def go_to_sfo():
...
+    """ 
+    Extends the class Node. 
+    self.action:     contains the optimistic description of an angelic HLA
+    self.action_pes: contains the pessimistic description of an angelic HLA
+    """
+
+    def __init__(self, state, parent=None, action_opt=None, action_pes=None, path_cost=0):
+        super().__init__(state, parent, action_opt, path_cost)
+        self.action_pes = action_pes 
+
```
delaray: Why is action_opt being set by the parent init method and action_pes being set by this init method, instead of both of them being set in this init method? Just curious.

tests/test_planning.py
```diff
@@ -283,17 +283,176 @@ def test_job_shop_problem():
...
+    assert(ang3.convert(ang3.effect) == [expr('PosNotA')])
+    assert(ang4.convert(ang4.effect) == [expr('PosYesNotA')])
+
+
+def test_is_primitive():
+    """
+    Tests if a plan is consisted out of primitive HLA's (angelic HLA's)
+    """
+    library = {
+        'HLA': ['Go(Home,SFO)', 'Go(Home,SFO)', 'Drive(Home, SFOLongTermParking)', 'Shuttle(SFOLongTermParking, SFO)', 'Taxi(Home, SFO)'],
```
delaray: Since several test functions use this same library. Define the library once (as a global variable) at the beginining of the angelic_search tests and share that definition across the different tests.

ad71:
@delaray the unit test functions have been temporarily fixed in #942 
The Travis build will pass once that is merged. There are no changes required to this pull request (atleast for this issue)

delaray:
@ad71 Awesome. Thank you.

planning.py
MariannaSpyrakou: I have added comments explaining the functionality of angelic_action, 
but maybe I need to add some more

planning.py
MariannaSpyrakou: I have commented on what should be the purpose of making_progress and what it is computing at the moment.

planning.py
MariannaSpyrakou: I just wanted to make full use of the parent object's variables, so I used the action variable to represent the optimistic description of the action. 
It is very easy for me to change it and add a variable action_opt.

delaray:
Angelic search code merged into master.

PR #945, opened 2018-08-08
Update README.md

tigerneil:
change to right directory


ad71:
You actually need to cd into aima-python, and running the next two commands generates the aima-data folder. It was correct the way it was before.

tigerneil:
sorry for that. it can be my mistake before.

PR #965, opened 2018-10-01
Resolving Issue 951 (Partial)

mkhalid1:
Error occurs while executing the *removed* line in the notebook; it has been replaced with a new line which gives the expected output. The solution is partial as another occurs later in the notebook, the solution to which is provided in my next commit in the mdp.py file.

### DIFF
--- a/mdp_apps.ipynb
+++ b/mdp_apps.ipynb
@@ -198,7 +198,7 @@
     "        actlist = list(set(actlist))\n",
     "        print(actlist)\n",
     "\n",
-    "        MDP.__init__(self, init, actlist, terminals=terminals, gamma=gamma)\n",
+    "        MDP.__init__(self, init, actlist, terminals, transition_matrix, rewards, gamma=gamma)\n",
     "        self.t = transition_matrix\n",
     "        self.reward = rewards\n",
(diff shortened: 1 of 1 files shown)


mdp_apps.ipynb
```diff
@@ -232,7 +232,7 @@
     }
    ],
    "source": [
-    "mdp = CustomMDP(t, rewards, terminals, init, gamma=.9)"
+    "MDP.__init__(self, init, actlist, terminals, transition_matrix, rewards, gamma=gamma)"
```
ad71: The point of the original line was to instantiate an object of the `CustomMDP` class and store it in the `mdp` variable.
Constructor calls should be within a derived class, and I believe your constructor call should replace line number 201 instead.

ad71: I don't think it would completely resolve the issue, but what you've done essentially avoids creating the `mdp` object and will cause problems when the `value_iteration`, `policy_iteration` and `best_policy` functions are later called, as all of them operate on the `mdp` object.

mkhalid1: True that. Changed Line 201 instead, so that an mdp object is still created, while resolving the error. @norvig please review

PR #966, opened 2018-10-01
Resolve Issue 951 (Completed)

mkhalid1:
The function "get_states_from_transitions" also needed to be modified to completely resolve the issue 951.

### DIFF
--- a/mdp.py
+++ b/mdp.py
@@ -78,6 +78,10 @@ class MDP:
     def get_states_from_transitions(self, transitions):
         if isinstance(transitions, dict):
             s1 = set(transitions.keys())
+            #s2 = set(tr[1] for actions in transitions.values()
+                     #for effects in actions.values()
+                     #for tr in effects)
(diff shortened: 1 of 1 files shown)


mkhalid1:
@ad71  The remaining issue i.e. the error we get when we call "value_iteration(mdp)" is in the lines of code that I have commented out (value of s2, basically). As if you remove them, the error disappears and the notebook executes quite nicely except that Class 2 is set wrongly to "study" instead of "sleep". Please check.

mkhalid1:
@ad71 Agreed, I'll look into it as well.

mkhalid1:
@norvig Please review as well; the two PRs (#965 and #966)  related to Issue 951 get the notebook working again, removing the errors and producing the expected output.

norvig:
I don't want to check in commented-out code. Let's really understand what is going on and get it right, not just get correct output on one example.

PR #967, opened 2018-10-02
Updated label_queen_conflicts function

mkhalid1:
Shortened it, finding conflicts separately and storing them in different variables has no use later in the notebook; so i believe this looks better. Plus, it makes the already lengthy notebook a tad bit shorter.


mkhalid1:
@ad71 Please review

PR #980, opened 2018-10-29
change in if else

hackerashish25:
there is no need of if-else statement 
if hidden_layer_size is zero then layer_size would be sum of only input_unit and output_unit


hackerashish25:
while loop in cross_validation_wrapper is fixed

PR #982, opened 2018-10-31
Updated CountingProbDist

adityauser:
If Observation 'o' is not initialized and increment is done using addition assignment, then dictionary will throw an error.


MrDupin:
Before the addition, we smooth for `o` (by calling `smooth_for(o)`), so `o` will be initialized in any case, unless I am missing something.

PR #986, opened 2018-11-29
Added new NPuzzle class Problem subclass(Generalization of Eight puzzle to 15-puzzle, 24-puzzle ect. )

SandersLin:
Generalization of Eight puzzle class to work with larger size sliding puzzles. 
I plan to replicate the results of branching factor comparison of various heuristics in AIMA book using 15-puzzle. I also implemented a Shuffling function to permutate the puzzle before solving.

### DIFF
--- a/search.py
+++ b/search.py
@@ -487,22 +487,17 @@ class EightPuzzle(Problem):
 
 # ______________________________________________________________________________
 class NPuzzle(Problem):
-    """ Generalization of  the Eight Puzzle problem. The problem  consists of sliding tiles numbered from 1 to N ^2  on a NxN board,
+    """Generalization of  the Eight Puzzle problem. The problem  consists of sliding tiles numbered from 1 to N ^2  on a NxN board,
     where one of the squares is a blank. The state is represented as a tuple of length N^2,
-    where element at index i represents the tile number at index i , where '0' denotes empty square index
-    """
-
-
-
-    def __init__(self, initial= None, goal= None, size = 3, heuristic = 'hamming', shuffle =10):
-        """   Args:
-                            intial: intiail state of puzzle. If not passed will default to goal state
-                            size: number of rows and columns in puzzle
-                            heuristic: heuristic used in astat search
-                            shuffle: number of times to perfrom random action from the initial state.
-                    Returns:
-                        Problem instance modeling any size of sliding puzzle
-                         """
+    where element at index i represents the tile number at index i , where '0' denotes empty square index"""
+    def __init__(self, initial=None, goal=None, size=3, heuristic='hamming', shuffle=10):
+        """Args:
+                intial: intiail state of puzzle. If not passed will default to goal state
+                size: number of rows and columns in puzzle
+                heuristic: heuristic used in astat search
+                shuffle: number of times to perfrom random action from the initial state.
+           Returns:
+                  Problem instance modeling any size of sliding puzzle"""
         self.size = size
         self.heuristic = heuristic
         goal = goal or (tuple(range(1,size*size, 1)) + (0, ))
@@ -513,8 +508,9 @@ class NPuzzle(Problem):
         assert len(initial) == size * size, "length of initial tuple should be {length}".format(length=size * size)
         assert self.check_solvability(initial), "The puzzle is not solvable from the given initial state!"
         Problem.__init__(self, initial, goal)
+        
     def shuffle_state(self, shuffle, state):
-        """ perform random action from the initial state"""
+        """Perform random action from the initial state"""
         for _ in range(shuffle):
             action = random.sample(self.actions(state),1)[0]
             state = self.result(state,action)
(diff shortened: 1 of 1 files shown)


SandersLin:
First opensource contrib ever! :) 
Please advise revisions.

search.py
```diff
@@ -484,6 +484,130 @@ def h(self, node):
...
+        goal = goal or (tuple(range(1,size*size, 1)) + (0, ))
+        initial = initial or (tuple(range(1,size*size, 1)) + (0, ))
+        if shuffle:
+            initial = self.shuffle_state(shuffle, list(initial))
+        assert len(goal) == size*size, "length of goal tuple should be {length}".format(length=size*size)
+        assert len(initial) == size * size, "length of initial tuple should be {length}".format(length=size * size)
+        assert self.check_solvability(initial), "The puzzle is not solvable from the given initial state!"
+        Problem.__init__(self, initial, goal)
+    def shuffle_state(self, shuffle, state):
+        """ perfrom random action from the initial state"""
```
kjacks21: type in the docstring here (perfrom -> perform)

search.py
```diff
@@ -484,6 +484,130 @@ def h(self, node):
...
+    def h(self, node):
+        """ Return the heuristic value for a given state. Default heuristic function used is
+        h(n) = number of misplaced tiles """
+        heuristics = {
+            'manhattan_distance':self.manhattan_distance_heuristic(node),
+            'misplaced_tiles':self.misplaced_tiles_heuristic(node)
+        }
+        return heuristics.get(self.heuristic)
+
+    def manhattan_distance_heuristic(self, node):
```
kjacks21: I'm not intimately familiar with the repo and the way the maintainers like things, but it may be worth calling the `manhattan_distance` and `hamming_distance` methods from [learning.py](https://github.com/aimacode/aima-python/blob/c16bb8bdc28e8f9fcc6e7f76a92b9492bf019d87/learning.py) to avoid reimplementing code already in the repo.

MrDupin: Good catch!

I would go even further and say that the distance functions should be moved to `utils.py`, since they are not directly tied to learning algorithms.

search.py
SandersLin: Thank you for spotting that! I'll correct that in the next commit.

search.py
SandersLin: Thank you for the suggestion! :)  I'm not familiar with the contribution guidelines as well. So I removed `manhattan_distance`  (the implementation in learning.py is applicable only to 1 dimension distance )  and used `hamming_distance` from `learning.py` isntead.

search.py
```diff
@@ -15,6 +15,7 @@
 import random
 import sys
 import bisect
+from .learning import hamming_distance
```
MrDupin: Not sure why you need the '.' since both files are in the same directory.

search.py
```diff
@@ -493,7 +494,7 @@ class NPuzzle(Problem):
 
 
 
-    def __init__(self, initial= None, goal= None, size = 3, heuristic = 'manhattan_distance', shuffle =10):
+    def __init__(self, initial= None, goal= None, size = 3, heuristic = 'hamming', shuffle =10):
```
MrDupin: Don't leave spaces around '='. For example, it should be `size=3` not `size = 3`.

search.py
```diff
@@ -513,25 +514,20 @@ def __init__(self, initial= None, goal= None, size = 3, heuristic = 'manhattan_d
         assert self.check_solvability(initial), "The puzzle is not solvable from the given initial state!"
         Problem.__init__(self, initial, goal)
     def shuffle_state(self, shuffle, state):
-        """ perfrom random action from the initial state"""
-
+        """ perform random action from the initial state"""
```
MrDupin: Comments should start with a capital letter. Also, comments should be all formatted the same way. Either leave a space around the comment (inside the string) from both sides, or keep it without spaces on the sides. Here you can either remove the starting space or add a space at the end of the sentence.

search.py
```diff
@@ -587,24 +576,24 @@ def check_solvability(self, state):
...
-
         return solvable
 
     def h(self, node):
         """ Return the heuristic value for a given state. Default heuristic function used is
         h(n) = number of misplaced tiles """
         heuristics = {
-            'manhattan_distance':self.manhattan_distance_heuristic(node),
-            'misplaced_tiles':self.misplaced_tiles_heuristic(node)
+
```
MrDupin: Don't leave this line blank.

search.py
```diff
@@ -587,24 +576,24 @@ def check_solvability(self, state):
...
         return heuristics.get(self.heuristic)
 
-    def manhattan_distance_heuristic(self, node):
 
-        return sum(abs(s%self.size - g%self.size) +
-                   abs(s//self.size - g//self.size)for (s, g) in zip(node.state, self.goal))
-    def misplaced_tiles_heuristic(self, node):
-        return sum(s != g for (s, g) in zip(node.state, self.goal))
+    def hamming_distance_heuristic(self, node):
+
```
MrDupin: Don't leave a blank line here.

antmarakis (changes requested):

MrDupin:
There are some stylistic issues in your comment, as I've noted above.

SandersLin:
Thank you so much for taking the to review my code. Didn't know I had so many stylistic issues before. I guess that's the benefit of open source lol.

SandersLin:
@MrDupin revisions made. Thanks

antmarakis (approved):

ad71 (approved):
Good work. Thanks!

dmeoli:
Thanks @SandersLin! Your fork branch was gone, so we rebased NPuzzle onto master and merged it (authorship preserved), with a test. We dropped the redundant 'from learning import hamming_distance' since utils.hamming_distance is already available in search.py. Closing in favor of the new PR.

Issue #890, opened 2018-03-25
Future Work on the NLP Apps Notebook


MrDupin:
Recently in the [nlp_apps](https://github.com/aimacode/aima-python/blob/master/nlp_apps.ipynb) notebook I added a section on the [Federalist Papers](https://en.wikipedia.org/wiki/The_Federalist_Papers). What I did was write a simple workflow from start to finish. There is a lot of work to be done still and I am opening this to community contributions, since I believe it is a great way to get started with the applications notebooks.

A few ways you can improve the section:

* One big issue with the Naive Bayes Classifier in this problem is that the multiplication of probabilities causes underflow (all the probability multiplications result in 0.0). That happens because examples are long texts. To avoid this, we are currently using the `decimal` module of Python. I believe this problem can be solved more elegantly using the logarithm of probabilities instead of probabilities. So instead of multiplying the probabilities, we add their logarithms.

* Do some pre-processing. Currently I only added a sample pre-processing step (removing one common word from each paper). I would like to see some other pre-processing tasks + analysis of the text. Which are the most common words for each author? Is it worth it if we removed the most popular words?

* Right now we are using unigram word models. There are other options available too. I would like to explore this in the notebooks. Maybe an author likes using two words together. Maybe another spells some words a bit differently. I would like to see different models used/explored in the notebook, to let the readers know that they shouldn't rely on just one model all the time. We can even combine models together.

* At the end of the notebook I note that the dataset is lopsided. We have way more information on Hamilton than the other two. Maybe it is worth adding some more writings from Jay/Madison to balance this out. I think it would be interesting to see if we could improve the results by using external data. This could come after the current section, so that we could compare the results.

* Finally, maybe we can take a step back and try and classify all the Federalist papers, not just the disputed ones. Add a new section where we use external data to train our model and then try and classify the papers.

---

This is a big undertaking, and it doesn't need to happen on the particular problem. If you have a problem in mind, you can instead use the above ideas to tackle your own problems! Sentiment Analysis is trending right now, so maybe this is a place to explore some of the above.

All in all, I think this is a good project to chip in every once in a while and I hope it will serve as an introduction to the repository. Or maybe it will sound interesting to GSoC students who might choose to tackle this.

In any case, feel free to post here with ideas + if you want to start working on something.

Dimkoim:
Hello @MrDupin! I have worked in NLP problems and I can give you some help here. :)

MrDupin:
[...]

PR #988, opened 2018-12-06
Add text pre-processing and visualization

Insiyaa:
Added stemmer and word cloud.
I didn't use the stemmed_papers as it lead to misclassification of 18-20 bt correctly classified 64. 
screenshot from 2018-12-06 23-41-22

Please review and let me know further changes.
This addresses #890.


nlp_apps.ipynb
```diff
@@ -559,6 +586,63 @@
     "len(jay), len(madison), len(hamilton)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's take a look at the common words used by them. <br>\n",
```
MrDupin: Here you need a newline after the end of the sentence to space things out.

nlp_apps.ipynb
```diff
@@ -559,6 +586,63 @@
...
...cZoWJNv6MY5NQIoNuKUJEmCrYujsxomjiIUXGU0OmByOthOhnCfgczlUVISRx4hF4vCUwHB+W1AlJFJymaKE1n3aVf9bCzFm7DRymN2/C3BZqlmTRHnx3ljS+s3lNh+acJ1cZlpsYfIZMqA0kXeP+eD1DMz7K2cZZGex5/sFbduI0Xd4Zfc+EnsU03qt9mpAudsDr4+YeP9D1P8+0Xkz1RysG6SmroQx9npSBcX9hSKI36XXaKuXbSAdot1ECd/nbc0nO63V86UzDp1kOkyXDdC5GwePL5H413u23nSadGhv+PVYTr17cd//2gtSIMN+NxIQSGtHHs+yfmYQg3drCc9LzkGNbWFNwWl67cTk0fPObWK1+7vhlDVjbeY6tDrVBRQBwFg599VLjZ/MxNHsB0sqTLM6RL0xiWg5nOkR6ZIug1yZRnaS6+g9dK2Avt1auM5UeBhKaeLs/gdaqM7H04sdM1LTKje5CGmVz/hKBXXaRXW9o22naLmp8b35+IWqazu/sIDTmIJ5KCCCKJR4VtDTVyhEjGPN2eYuGKz6PPZrl4xiUaKLa7fUWv/VdnK/ljlZirSBH2Q8J+SOwnc9TK1fjLy6iSm1x9doA0LNKFKfLlvZip3NBXcSesX3+ZoL+98qYVhP2IldfXqV6qM358lCO/cIB9H5yluDfPw79xjCO/cICNC3WufGmexZdXCHtbN7CYiHac0H0lSSApuUXl1cNZo1se6hKJQKKIkRiE+t0nHn7cJVYhUkg0EOsweX4hEyrw4EseeAm9SqMT4YXbzlOrHrO2GFBdjVhblKAhk5eMTlqs3gyIoiS4MAzoNGNMS1CeMHF7yXN7wXZBBT1892zZHG7dbggLQ1oIBLGOiFSAIUwMaSX0ZeUn51OYyXvTemAvpDGlgxQGSsfJ45CJMMhgtj2IXaSQmDLxb4xUgNIR3aBKtX8D20jzIHDfuYo9N4V/fZmo2kDYFtknTuB2KrjdrV0/rRSR28NwUsRun5GT70Oam183c7SEUS4T3FjY0b5vCEMiTAPt7z5YEoaBPTWFVS6TeThRQteuS7C+jtCQSgv6g8/MKOUpfuxx6p/5LmpQMY4jn159BxG/XULHMc3zr5GenCN/4DgzH/81gsYGdnEUYVqUH36a4pGHsUfGsUdGidwe1de/TdDe2gn3ejXyo/uIQneoyC5SDsKyEsZEkI
```
MrDupin: Replace the '&' with 'and' and remove the 'the' from 'Hamilton seems to use the words'. Also, when you are referencing a word put it in quotes. Like this:

They all have common words like "state", "will" and "government".

Finally, the final sentence ('Word clouds reveal some sort of sentiments if analyzed properly.') is a bit too vague. You need to add some more details, right now it doesn't carry much information.

antmarakis (changes requested):
I like the addition, but I noticed a few things I would like fixed.

Before I go into the review, I believe that since the stemmer is not actually being used for training, we should put it after the main part of the notebook, as an addendum. It's more of a bonus thing, so it shouldn't be in the middle of things. The word cloud analysis looks fine where it is. Generally we do not like using external libraries, but for the case of visualization there is not much that can be done, unless `matplotlib` has something for word clouds.

nlp_apps.ipynb
```diff
@@ -479,7 +479,27 @@
...
     "\n",
-    "NOTE: Since we are only removing a single word from each paper, this step can be skipped. We add it here to show that processing the data in our hands is something we should always be considering. Oftentimes pre-processing the data in just the right way is the difference between a robust model and a flimsy one."
+    "NOTE: Since we are only removing a single word from each paper, this step can be skipped. We add it here to show that processing the data in our hands is something we should always be considering. Oftentimes pre-processing the data in just the right way is the difference between a robust model and a flimsy one. \n",
+    "\n",
+    "For example,after getting the words, we need the root form rather than the original word. For eg. the words \"support\", \"supportive\", \"supporting\" have the root word \"support\". It is better to have a single root word for multiple different words with similar meaning. It reduces the dimension of data. There are many stemming algorithms in NLP, popular ones being the porter stemmer and snowball stemmer.\n",
```
MrDupin: This paragraph needs a bit of work:

a) Add a space after the first comma. Remember that after punctuation you need to have a space (at least, most of the time).

b) "For eg." - Write the word 'example' out instead of 'eg.'

c) "It reduces the dimension of data" - The word 'dimension' needs to be plural, I believe.

d) "porter stemmer and snowball stemmer" - Algorithm names need to be capitalized.

nlp_apps.ipynb
```diff
@@ -572,7 +559,7 @@
    "source": [
     "disputed = list(range(49, 58+1)) + [18, 19, 20, 64]\n",
     "jay, madison, hamilton = [], [], []\n",
-    "for i, p in enumerate(papers):\n",
+    "for i, p in enumerate(stemmed_papers):\n",
```
MrDupin: You have not defined `stemmed_papers`.

As I previously suggested, it would be better if you added the stemming section after the original code, to compare the two.

Insiyaa:
Thanks for your patience! :).. Please review...

nlp_apps.ipynb
```diff
@@ -860,6 +840,141 @@
...
+    "    probs = recognize(papers[d], nBSL)\n",
+    "    results = ['{}: {:.6f}'.format(name, probs[(name, 1)]) for name in 'Hamilton Madison Jay'.split()]\n",
+    "    print('Paper No. {}: {}'.format(d, ' '.join(results)))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The stemmer didn't seem to help much here. But the model classifying Madison's papers as Jay's after stemming the words may mean that they used the words having same root words. The classifier could perform well if we had more instances of their work."
```
MrDupin: This looks better, nice work! I noticed a little issue with the last paragraph though. You write this:

"stemming the words may mean that they used the words having same root words."

First of all, you use the word 'words' three times in quick succession. This should generally be avoided. Second, the sentence needs to be made a bit clearer. Right now it is hard to understand what you mean.

antmarakis (approved):

dmeoli:
Thanks @Insiyaa. Stemming and word clouds pull in extra dependencies (e.g. wordcloud) and live mostly as notebook output, which is outside the core library scope. Closing for now; thanks for the contribution!

PR #989, opened 2018-12-15
pre-processing of text

hackerashish25:


MrDupin:
Thanks for the contributions, much appreciated!

Unfortunately I will have to close this PR, since you perform many changes that could be broken into multiple PRs. For any contribution, we encourage users to focus on one thing to keep things simple. Here you do three things in one PR, which is not ideal. Please break it up into multiple PRs, thank you.

Issue #896, opened 2018-03-27
Invalid syntax in gui/eight_puzzle.py


LoboLofi:
I'm trying to run the Gui examples, with:
`python3 -i gui/eight_puzzle.py`
And get the next out:

```
File "gui/eight_puzzle.py", line 88
    b[zero_ix] = Button(root, text=f'{state[index]}', width=6, font=('Helvetica', 40, 'bold'), command=partial(exchange, zero_ix))
                                                   ^
SyntaxError: invalid syntax
```

I've python 3.5.2 and Ubuntu 16.04.

apb7:
Hi @LoboLofi, this isn't a syntax error actually. This feature has been introduced in Python 3.6 and would run smoothly if you update your Python version. Thanks!

LoboLofi:
Thanks a lot @apb7

sagar-sehgal:
@apb7 @ad71 It has been mentioned in the [README](https://github.com/aimacode/aima-python#python-34-and-up) that we can run it with Python3.4 and up.  
So we can still make this compatible for lower version of python by using 
`b[zero_ix] = Button(root, text=state[index], width=6, font=('Helvetica', 40, 'bold'), command=partial(exchange, zero_ix))`
There is a simmilar isssue in [gui/grid_mdp.py](https://github.com/aimacode/aima-python/blob/master/gui/grid_mdp.py) .
This code is running just fine on my system after this change which has simmilar configurations as mentioned above.
Shall I go ahead and create a PR for this? Or is there any gap in my solution?

ad71:
@sagar-sehgal As long as everything still works on Python 3.4+, we're good. You can submit a PR.

sagar-sehgal:
@ad71 @MrDupin  Sorry for replying late but, I submitted th PR for the changes in gui/eight_puzzle.py and gui/genetic_algorithm_example.py  but I didn't do that for gui/grid_mdp.py . Actually  on applying the above changes in grid_mdp.py and then running the code, it was fine until I clicked on a cell. On clicking on a cell and providing it a reward/making it a terminal/making it a cell it was showing a new error
```
Exception in Tkinter callback
Traceback (most recent call last):
  File "/usr/lib/python3.5/tkinter/__init__.py", line 1553, in __call__
    return self.func(*args)
  File "grid_mdp.py", line 128, in update_table
    if wall.get() == WALL_VALUE:
  File "/usr/lib/python3.5/tkinter/__init__.py", line 358, in get
[...]

PR #991, opened 2018-12-16
Solved the Issue#896 in gui/eight_puzzle.py and gui/genetic_algorithm…

thesagarsehgal:
Solved the Issue #896 in the filesgui/eight_puzzle.py and gui/genetic_algorithm_example.py . These files were only compatible with Python3.6+ and throwed and error with Python3.4 and Python3.5.

### DIFF
--- a/gui/eight_puzzle.py
+++ b/gui/eight_puzzle.py
@@ -85,7 +85,7 @@ def exchange(index):
 
 	if current_action in actions:
 		b[zero_ix].grid_forget()
-		b[zero_ix] = Button(root, text=f'{state[index]}', width=6, font=('Helvetica', 40, 'bold'), command=partial(exchange, zero_ix))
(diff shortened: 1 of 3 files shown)


gui/eight_puzzle.py
```diff
@@ -85,7 +85,7 @@ def exchange(index):
 
 	if current_action in actions:
 		b[zero_ix].grid_forget()
-		b[zero_ix] = Button(root, text=f'{state[index]}', width=6, font=('Helvetica', 40, 'bold'), command=partial(exchange, zero_ix))
```
sagar-sehgal: f'{var}' was introduced in Python3.6 only

ad71 (approved):

dmeoli:
Thanks @thesagarsehgal. Issue #896 was about f-strings on older Python; the repo now targets Python 3.9+, where f-strings work fine, so converting them to str() isn't needed (and it's a bit less idiomatic). Closing as no longer applicable — thanks!

Issue #948, opened 2018-08-11
What I didn't do in GSoC 2018


ad71:
__What I _did_ this summer:__
[Here is a gist](https://gist.github.com/ad71/7b3596e711b0e66d76cbae1738347b6f) of what I did. It's just an enumeration of the modules/algorithms that I worked on and doesn't go into much detail. In case someone has queries or questions, hit me up anytime.

__What _remains_ to be done:__
Everything apart from the following three algorithms, has been implemented with tests and documentation.
- The [`HybridWumpusAgent`](https://github.com/aimacode/aima-python/blob/master/logic.py#L1009) class needs fixing. Some helper functions don't work as expected. The algorithm has no tests and no documentation. Some related helper functions take too long to run and require refactors or re-implementation.
- The [`cross_validation`](https://github.com/aimacode/aima-python/blob/master/learning.py#L1035) section needs closure. Refer to [this issue](https://github.com/aimacode/aima-pseudocode/issues/55), [this issue](https://github.com/aimacode/aima-python/issues/566) and all linked issues for more details. Tests and documentation is incomplete.
- The [`DecisionListLearner`](https://github.com/aimacode/aima-python/blob/master/learning.py#L621) is not yet complete. Refer to [this issue](https://github.com/aimacode/aima-python/issues/856). Tests and documentation is incomplete as well.

__What _else_ can be done:__
- The project could benefit from GUI apps for some algorithms or standalone visualizations which users can interact with. [Here](https://github.com/aimacode/aima-python/tree/master/gui) are some examples built using `tkinter`. I made a couple of these, but quite a few algorithms need a corresponding applet.
- Some algorithms need better and more intuitive explanations in the notebooks. Several notebook sections are pretty drab, and pictures, visualizations, graphs or interactive widgets could help.
- Various algorithms from the `learning.py` module cannot handle multi-dimensional data (like images) well. If this can be improved, the module can be used as a neural network and pre-processing library for simple tasks and will increase the utility of the project as a whole.
- NLP enthusiasts can try [this](https://github.com/aimacode/aima-python/issues/890).
- Also, please refer to the _what I didn't do_ sections of @MrDupin's [writeup](https://github.com/aimacode/aima-python/issues/632) and @Chipe1's [writeup](https://github.com/aimacode/aima-python/issues/641) from last year. Some of it still requires attention.

[...]

PR #998, opened 2018-12-21
Decision List Learner implemented

pyaf:
Hi,

The implementation of DecisionListLearner was incomplete (discussed in #948). I've implemented the algorithm, written tests and updated the corresponding notebook.

Here's what I've done:

* [DLTest](https://github.com/pyaf/aima-python/blob/DLTestLearner/learning.py#L622) class represents tests in the decision list. A `DLTest` has `attr_values` to hold attributes and corresponding values to test them against. It has some member functions like calling `matches` with an example as input checks whether the example passes the test or not, calling `matched_examples` or `unmached_examples` with inputs as a list of examples returns a subset of examples which pass and don't pass the test respectively.

* [create_DLTests](https://github.com/pyaf/aima-python/blob/DLTestLearner/learning.py#L654) creates a list of all possible `DLTest`s which are used in creating Decision List. (Decision List Learner repeatedly loops on these tests so as to find a test that agrees exactly with some subset of the training set.) It takes as input `attribute_count` which is the maximum number of attributes to consider while creating a test. As of now `attribute_count`=1 is supported.

* [DecisionListLearner](https://github.com/pyaf/aima-python/blob/DLTestLearner/learning.py#L672) is the implementation of the of Decision list learner as mentioned in the book. Member function `decision_list_learning` is the python implementation of the [pseudocode](https://github.com/aimacode/aima-pseudocode/blob/master/md/Decision-List-Learning.md)

* I've added test for `DecisionListLearner` [here](https://github.com/pyaf/aima-python/blob/DLTestLearner/tests/test_learning.py#L178).

* I've updated the [learning.ipynb](https://github.com/pyaf/aima-python/blob/DLTestLearner/learning.ipynb) to include `DecisionListLearner`. I've also added an [image](https://github.com/pyaf/aima-python/blob/DLTestLearner/images/decision_list.png) in its description. 

The code is well documented, let me know if you need any clarifications regarding the code.


ad71 (approved):
Great work! Thanks!

pyaf:
Any updates on this PR?

learning.py
```diff
@@ -618,38 +618,114 @@ def predict(example):
...
-# A decision list is implemented as a list of (test, value) pairs.
 
+class DLTest:
+    def __init__(self):
+        self.attr_values = {}
 
-def DecisionListLearner(dataset):
-    """[Figure 18.11]"""
+    def add(self, attr, value):
+        """add new attribute and value to the test"""
```
MrDupin: Minor edit, but can you please capitalize the start of your sentences, for consistency's case?

antmarakis (commented):
Overall it seems like a good PR, but you should wait for @norvig to respond, since it is too substantial for me to merge.

learning.py
pyaf: Sure!

dmeoli:
Thanks @pyaf. You're right that DecisionListLearner is an incomplete stub on master (find_examples/passes raise NotImplementedError). However, this implementation doesn't conform to the standard learner interface (DecisionListLearner(dataset) should return a callable predictor; here it isn't callable), is limited to single-attribute tests, and is a large 2019 rework. Completing DecisionListLearner properly (interface-conforming + tested) is worth doing as a fresh focused PR. Closing this version — thanks for flagging #948!

PR #1001, opened 2018-12-25
Grammar and typo fixes in logic notebook

dsaw:


MrDupin:
I'm sorry, this has too many commits. Break it down into smaller ones so that we can properly review them, thanks.

PR #1006, opened 2019-01-12
Implementing Vectorization

glitch401:
Vectorization can reduce the run time by  replacing large loop operations significantly !

For proof of Run time improvements 
image

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -132,7 +132,7 @@ def histogram(values, mode=0, bin_function=None):
 
 def dotproduct(X, Y):
     """Return the sum of the element-wise product of vectors X and Y."""
-    return sum(x * y for x, y in zip(X, Y))
+    return np.dot(X,Y)
 
 
 def element_wise_product(X, Y):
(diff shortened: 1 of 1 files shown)


ad71 (approved):
Nice catch, thanks!

indranilbiswas7:
the performance of runtime significantly increases using vectorization!

MrDupin:
Even though this is quite an improvement, I don't know if this is needed. Is there a place in the repository that would benefit from this change? Some function that takes too long because of the current implementation, perhaps?

I would prefer to keep the code as "bare-bones" as possible, without the use of external functionality, unless absolutely necessary.

indranilbiswas7:
I would totally agree with your perspective of keeping the code to the bare bones, Sir. But I think, the dot product is used multiple times throughout the repo, so just was a suggestion to improve the runtime, its only an if statement, with 4 lines of code added in total. :)

dmeoli:
Thanks @glitch401. This swaps dotproduct to np.dot when lengths match; it changes the return type to a numpy scalar, has minor style issues, and the unequal-length branch isn't meaningful for a dot product. Marginal vs the current code. Closing — thanks!

PR #1011, opened 2019-01-22
necessary change in learning.py file 

hackerashish25:
removal of unnecessary if else statement


learning.py
```diff
@@ -839,10 +839,8 @@ def network(input_units, hidden_layer_sizes, output_units, activation=sigmoid):
     excluding input and output layers
     """
     # Check for PerceptronLearner
```
MrDupin: Please remove this comment and the new line below it.

hackerashish25: @MrDupin  okay . Thank you for review

PR #1014, opened 2019-01-22
update in multimap function in utils.py and added test for it

sagar-sehgal:
updated the multimap function and added test cases to test it.


tests/test_utils.py
```diff
@@ -25,6 +25,9 @@ def test_count():
     assert count([True, False, True, True, False]) == 3
     assert count([5 > 1, len("abc") == 3, 3+1 == 5]) == 2
 
+def test_multimap():
+    assert multimap([(1,2),(1,3),(1,4),(2,3),(2,4),(4,5)]) == {1:[2,3,4],2:[3,4],4:[5]}
```
MrDupin: Please add some spaces between commas and colons ( ':' ). Also, if the line gets too long, you may need to split it into two.

tests/test_utils.py
```diff
@@ -25,6 +25,9 @@ def test_count():
...
+    assert multimap([("a",2),("a",3),("a",4),("b",3),("b",4),("c",5)]) == {'a': [2, 3, 4], 'b': [3, 4], 'c': [5]}
```
MrDupin: Same as above.

PR #1016, opened 2019-01-23
Added some test cases to first function in utils.py

sagar-sehgal:
Added more test cases to first function

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -35,9 +35,14 @@ def test_first():
     assert first('word') == 'w'
     assert first('') is None
     assert first('', 'empty') == 'empty'
+    assert first([1,2,3,4,5]) == 1
+    assert first([]) == None
     assert first(range(10)) == 0
     assert first(x for x in range(10) if x > 3) == 4
     assert first(x for x in range(10) if x > 100) is None
+    assert first((1,2,3)) == 1
+    assert first([(1,2),(1,3),(1,4)]) == (1,2)
+    assert first({1:"one",2:"two",3:"three"}) == 1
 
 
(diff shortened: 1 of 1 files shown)


tests/test_utils.py
```diff
@@ -35,9 +35,14 @@ def test_first():
     assert first('word') == 'w'
     assert first('') is None
     assert first('', 'empty') == 'empty'
+    assert first([1,2,3,4,5]) == 1
```
MrDupin: To conform with PEP8, you should add a space after each comma.

tests/test_utils.py
```diff
@@ -35,9 +35,14 @@ def test_first():
...
+    assert first([]) == None
     assert first(range(10)) == 0
     assert first(x for x in range(10) if x > 3) == 4
     assert first(x for x in range(10) if x > 100) is None
+    assert first((1,2,3)) == 1
```
MrDupin: Here as well.

tests/test_utils.py
```diff
@@ -35,9 +35,14 @@ def test_first():
...
+    assert first([(1,2),(1,3),(1,4)]) == (1,2)
```
MrDupin: And here.

tests/test_utils.py
```diff
@@ -35,9 +35,14 @@ def test_first():
...
+    assert first({1:"one",2:"two",3:"three"}) == 1
```
MrDupin: And here.

tests/test_utils.py
sagar-sehgal: Ok @MrDupin , I'll do that .

PR #1018, opened 2019-01-25
cross_validation_wrapper loop is fixed

ashishgit7:
forever while loop is fixed and it return learner

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -1102,9 +1102,11 @@ def cross_validation_wrapper(learner, dataset, k=10, trials=1):
     err_val = []
     err_train = []
     size = 1
-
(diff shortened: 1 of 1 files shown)


hackerashish25:
@ad71 can you review this PR

SiluPanda:
Why is that, once the validation error is greater than the previous validation error, the function converges? Is there any proof (mathematical, if possible) of that?

dmeoli:
Thanks @ashishgit7. cross_validation_wrapper has since been reimplemented in the learning.py refactor, so this 2019 loop fix no longer applies (same as #1052). Closing as superseded — thanks!

Issue #1015, opened 2019-01-23
product function in `utils.py`


sagar-sehgal:
In `utils.py` file in the [`product`](https://github.com/aimacode/aima-python/blob/25c2a9b6ef861a018554318e6c1a5741924a425a/utils.py#L56) function .
But if we call functions like 
- `product([1,"a",3])`  return "aaa"
This is correct but what should I do with it. Should a test be added for it?
- `product([1,"a","b",3])` gives error for multiplication of 2 strings.
Or an assertion must be created to check this at the beginning of the function.

hackerashish25:
@sagar-sehgal  product([1,"a",3]) is returning "aaa" so how it could be correct . [product](https://github.com/aimacode/aima-python/blob/25c2a9b6ef861a018554318e6c1a5741924a425a/utils.py#L56) shold return product of number otherwise assert an error

sagar-sehgal:
I tried to solve the issue by adding an assert statement like:-
```    
for num in numbers:
        assert (isinstance(num, (int, float))), "The variable in the list is not a number"
```
it solved the above issue but on testing 2 cases failed.

![Screenshot from 2019-03-09 22-33-10](https://user-images.githubusercontent.com/22503629/54074669-55f5cb00-42bb-11e9-93b6-b3bfed4a365b.png)

The reason being that, for multiplying, instead of sending the list, it is being called from a generator like [here](https://github.com/aimacode/aima-python/blob/a4c4c4609a458cac4b039ca9e2123a77f876b893/learning.py#L340) and [here](https://github.com/aimacode/aima-python/blob/a4c4c4609a458cac4b039ca9e2123a77f876b893/learning.py#L368)

MrDupin:
I don't think this is an issue, more of a side-effect. `product([1,"a",3])` is not something that we are expecting to do here, so there is no need for any fix. The function is used to compute the product in a list of numbers, so there is no need to perform any checks.

Remember that this is an educational project that aims at teaching AI. We don't have to be this careful about this sort of things. Readable code is better than catching edge-cases all the time.

PR #1020, opened 2019-01-25
Fixed Issue #1015

alannair:
Added code to `product` function in `utils.py` to check whether arguments are numbers (or strings that can be treated as numbers eg. 'one' ). Raises ValueError if otherwise


dmeoli:
Thanks @alannair. product() naturally raises on non-numeric inputs already, so this defensive type-check adds little; the change is also from 2019 against an older utils.py. Closing as minor; thanks for looking into #1015!

PR #1021, opened 2019-01-27
some test cases are added

ashishgit7:
more test cases in [test_utils.py](https://github.com/aimacode/aima-python/blob/master/tests/test_utils.py)


dmeoli:
Thanks @ashishgit7! This is a small 2019 change (and uses tab indentation / touches several files); it's marginal against the current code. A focused, PEP 8 PR would be welcome if still relevant. Closing for now — thanks!

Issue #1030, opened 2019-02-19
Import statements in try-except block


rajatjain1997:
In `agents.py` I stumbled onto the following import statements:

https://github.com/aimacode/aima-python/blob/e9033cf7a46478a8af93831784695a68c128e165/agents.py#L577-L582

These import statements are used to render graphics by the `GraphicEnvironment` class defined just below it. However, since these packages are not mentioned in `requirements.txt`, neither binder nor any developer installs them before running the notebooks. This results in the following stacktrace when `agents.ipynb` tries to create an instance of `GraphicEnvironment`:

```

---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-18-0a21d3807079> in <module>
----> 1 park = GraphicPark(5,5, color={'EnergeticBlindDog': (200,0,0), 'Water': (0, 200, 200), 'Food': (230, 115, 40)})
      2 dog = EnergeticBlindDog(program)
      3 dogfood = Food()
      4 water = Water()
      5 park.add_thing(dog, [0,0])

/spare/aimacode/agents.py in __init__(self, width, height, boundary, color, display)
    588         but initialise a BlockGrid for GUI too."""
    589         super().__init__(width, height)
--> 590         self.grid = BlockGrid(width, height, fill=(200, 200, 200))
    591         if display:
    592             self.grid.show()

NameError: name 'BlockGrid' is not defined 
```
There are two ways to resolve this issue:
 - Remove the statements from the try except block and include these packages in `requirements.txt`
 - Put the entire class `GraphicEnvironment` in try-except so that it is only defined if those packages are installed.

I prefer the first approach since it requires minimal changes and because the second one will lead to the class not working on binder at all. However since the contribution guide states that no new packages should be added to `requirements.txt`, I'm at an impasse.

**Steps to reproduce the issue:**
 - Open Binder > agents.ipynb
 - Rerun all cells by restarting the kernel (This is required because the initial output is a cached version by the notebook)

dshahid380:
I would like to take this issue.

rajatjain1997:
Umm... My PR already works on solving the issue.

dshahid380:
But you did not get murge right.

ad71:
#1031 will be merged soon @dshahid380 .
This issue will be closed as soon as that is done. Till then, consider this resolved.

gamechanger98:
i would like to work on this issue as I am also having it.

rajatjain1997:
[...]

PR #1031, opened 2019-02-24
Rework agents.ipynb

rajatjain1997:
This PR contains improvements related to `agents.ipynb` notebook.

- Made changes to `agents.ipynb` to improve its reading flow.
- Re-implemented a wrong implementation of `EnergeticBlindDog` where the agent program could see the state of the agent by defining a global variable.
- Merged non-graphic implementation of `Park2D` with the graphics one.
- Fixed issue #1030 using the second approach.

**Note:** `agents.ipynb` has been stripped of its output cells before being commited to git. This is because jupyter notebooks cache all outputs (including execution count) in the `ipynb` file. This leads to a few problems while working with any VCS:
 - Merge conflicts occur even when two authors work on separate sections of the same file (since execution count or python version are bound to be different)
 - It is tough to detect errors (See #1030 )
 - PR review process becomes difficult since diff becomes unnecessarily bloated with output details.


MrDupin:
I like the new structure. Didn't notice any major issues when looking at the ipynb code (haven't compiled it though).

Looks good!

rajatjain1997:
Thanks for the review. I'm sure that the code compiles and runs. Should we modify the contribution guide to ask contributors to strip all outputs from notebooks before they commit?

**P.S.:** `nbstripoutput` is an amazing utility for this.

dshahid380:
I would like to fix this issue.

MrDupin:
Output should be shown wherever possible. A lot of students may want to just read the notebooks online at GitHub. The rule of thumb is, show the output unless it is not convenient to do so (for example, when you are trying to print a very long list; in that case it would be best if only part of it was printed).

rajatjain1997:
That's a great point. However, since we have binder setup for online viewing, I thought we wanted to encourage students to run code as much as possible. But yeah, looking at output on GitHub is super-convinenent and may be something that a lot of students do.

Issue #1035, opened 2019-03-04
Giving manual choice for the agent location in xy_vacuum_environment.py


KanakalathaVemuru:
Hi!
I have seen your code and it seems to me that you can improve it by adding an additional feature of having the agent's position, flexible in xy_vaccum_environment. This can be done by adding an extra button to change the position of the agent. I have written and implemented the code and this is the screenshot of it. 
![vacuum](https://user-images.githubusercontent.com/46847239/53732818-0a9b8100-3ea5-11e9-8eb5-722f1663fc6d.jpg)
Hope it helps.

MrDupin:
This looks good, have you made a Pull Request with your changes?

KanakalathaVemuru:
Yes, I made a pull request.

MrDupin:
Looks good. Now you just have to wait for @norvig to merge this. Nice work!

dmeoli:
Closing as stale: the companion PR #1036 (same title) was closed without merge, this issue is several years old, and it concerns only the standalone `gui/xy_vacuum_environment.py` demo (not the core library). The change itself is small (let the user click a cell to place the agent instead of the hardcoded (3, 3)); if still wanted it can be reopened as a focused GUI PR.

PR #1036, opened 2019-03-05
Giving manual choice for the agent location in xy_vacuum_environment.py.

KanakalathaVemuru:
This PR is related to issue #1035.
I have added the manual choice for the agent in the file.
I have added the coordinate system for the buttons.


dmeoli:
Thanks @KanakalathaVemuru. The manual-agent-placement idea (#1035) is nice, but this implementation tracks state via a module-level global (agnt) instead of an instance attribute, which is fragile. Reworking it to per-instance state would be needed. Closing for now; a focused, instance-state version would be welcome. Thanks!

PR #1037, opened 2019-03-07
included removing sparse values from dictionnary

0xrushi:
@MrDupin please have a look at this

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -298,6 +298,18 @@ class CountingProbDist:
             self.sampler = weighted_sampler(list(self.dictionary.keys()),
                                             list(self.dictionary.values()))
         return self.sampler()
+        
+    def remove_sparse(self,sparse_value):
+    	""" keep items with probablity sparse value 
+    	and remove all others , eg if sparse_value=0.98 items with probablity
+    	0.02 would be removed"""
+    	prob_to_remove= 1-sparse_value
+    	tot = self.n_obs
+    	for (k, v) in self.dictionary.items():
+    		if v/tot < prob_to_remove:
+    			del self.dictionary[k]
+    			self.n_obs -= v
+
 
 # ______________________________________________________________________________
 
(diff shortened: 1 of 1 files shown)


learning.py
```diff
@@ -298,6 +298,18 @@ def sample(self):
             self.sampler = weighted_sampler(list(self.dictionary.keys()),
                                             list(self.dictionary.values()))
         return self.sampler()
+        
+    def remove_sparse(self,sparse_value):
+    	""" keep items with probablity sparse value
```
MrDupin: Capitalize comment start. Also, it should be 'Keep items with probability greater than sparse value' or something like that.

learning.py
```diff
@@ -298,6 +298,18 @@ def sample(self):
             self.sampler = weighted_sampler(list(self.dictionary.keys()),
                                             list(self.dictionary.values()))
         return self.sampler()
+        
+    def remove_sparse(self,sparse_value):
```
MrDupin: Add space after comma.

learning.py
```diff
@@ -298,6 +298,18 @@ def sample(self):
...
+    	""" keep items with probablity sparse value 
+    	and remove all others , eg if sparse_value=0.98 items with probablity
```
MrDupin: Remove space before comma and add period after 'eg'.

learning.py
```diff
@@ -298,6 +298,18 @@ def sample(self):
...
+    	0.02 would be removed"""
```
MrDupin: You are not only removing items with probability 0.02, you are removing items with probability less than 0.02. Also, add a period at the end of the comment.

antmarakis (commented):
I am not sure where or how this will be used, so I am not sure if we need this. Can you provide an example of this? Generally we try to only add code that has some use (either educational or implemention-wise). If this simply for educational purposes, you should add a bit of a description.

dmeoli:
Thanks @0xrushi. This remove_sparse method deletes from self.dictionary while iterating over self.dictionary.items(), which raises 'dictionary changed size during iteration', and uses tab indentation. Closing; a fixed, PEP 8 version (iterate over a list copy of the keys) in a focused PR would be welcome. Thanks!

PR #1039, opened 2019-03-07
Install dependencies from `requirements.txt` missing in `README.md`

sagar-sehgal:
An important step was missing in the `README.md`, i.e. the instructions to install the dependencies from `requirements.txt` in the system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ To download the repository:
 
 `git clone https://github.com/aimacode/aima-python.git`
 
+Then you need to install the basic dependencies to run the project in your system (although they are very less):
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,6 +31,10 @@ To download the repository:
 
 `git clone https://github.com/aimacode/aima-python.git`
 
+Then you need to install the basic dependencies to run the project in your system (although they are very less):
```
MrDupin: I'm not sure what you mean by 'although they are very less'. Is it a typo? Do you mean to say 'even though there are not a lot'? If that's the case, I don't think the parenthesis is needed.

PR #1040, opened 2019-03-08
Added remove_sparse()

kyscg:


kyscg:
Review @MrDupin ?

MrDupin:
The same as: https://github.com/aimacode/aima-python/pull/1037

PR #1042, opened 2019-03-11
Added Soundex - Phonetic Search Algorithm to `nlp_apps.ipynb`

thesagarsehgal:
Added the Soundex Algorithm to `nlp_apps.ipynb` as mentioned in https://github.com/aimacode/aima-python/issues/890#issuecomment-471327141 .

### DIFF
--- a/nlp_apps.ipynb
+++ b/nlp_apps.ipynb
@@ -17,7 +17,8 @@
     "\n",
     "* Language Recognition\n",
     "* Author Recognition\n",
-    "* The Federalist Papers"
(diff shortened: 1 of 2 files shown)


dmeoli:
Thanks @thesagarsehgal. Soundex is a phonetic-matching algorithm that's outside the scope of the AIMA text module (n-gram models / ciphers), and this implementation is also incomplete (it doesn't normalize to the standard 4-character code). Closing as out of scope; thanks for the contribution!

PR #1051, opened 2019-03-24
changed queue to set in AC3

DonatoMeoli:
Changed queue to set in AC3 (as in the pseudocode of the original algorithm) to reduce the number of consistency-check due to the redundancy of the same arcs in queue. For example, on the harder1 configuration of the Sudoku CSP the number consistency-check has been reduced from 40464 to 12562!

### DIFF
--- a/tests/test_csp.py
+++ b/tests/test_csp.py
@@ -13,7 +13,7 @@ def test_csp_assign():
     assignment = {}
     australia.assign(var, val, assignment)
 
-    assert australia.nassigns == 1
+    # assert australia.nassigns == 1
     assert assignment[var] == val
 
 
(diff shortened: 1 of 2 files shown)


tests/test_csp.py
```diff
@@ -13,7 +13,7 @@ def test_csp_assign():
     assignment = {}
     australia.assign(var, val, assignment)
 
-    assert australia.nassigns == 1
+    # assert australia.nassigns == 1
```
MrDupin: Why was this commented out? I am not familiar with AC-3, but generally removing tests is not good.

DonatoMeoli: I'm sorry, I added the test again! :+1:

PR #1052, opened 2019-03-25
fixed infinite while loop in cross_validation_wrapper

thevarunsharma:
I noticed that this problem was marked as TODO in the learning.py code, so I tried fixing it. I hope it will be working fine now i.e. no more forever running while loop.

### DIFF
--- a/learning.py
+++ b/learning.py
@@ -102,7 +102,7 @@ class DataSet:
         else:
             self.examples = examples
 
-        # Attrs are the indices of examples, unless otherwise stated.   
+        # Attrs are the indices of examples, unless otherwise stated.
         if self.examples is not None and attrs is None:
             attrs = list(range(len(self.examples[0])))
 
@@ -646,7 +646,7 @@ def DecisionListLearner(dataset):
         for test, outcome in predict.decision_list:
             if passes(example, test):
                 return outcome
-    
+
     predict.decision_list = decision_list_learning(set(dataset.examples))
 
     return predict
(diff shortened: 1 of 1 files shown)


SiluPanda:
Hey, what I see you are doing is, converging the training error to 0. But in that case, it would just be overfit on the training set, loosing the generality. For example:, let's say we have 4 data sets and four labels. The model might take a polynomial of higher degree (greater than 4) and easily classify all four of them correctly, but that would perform horribly on the validation set and test set.

I hope I'm making sense to you.

thevarunsharma:
Thanks for your suggestion @SiluPanda, but I want to clear thet after checkin convergence of training error, I am selecting the size which yielded the best val. error, so, it won't be choosing the overfitted solution.
Plus, please suggest a good threshold for train error, if close to 0 is not a good one.

SiluPanda:
There are a lot of ways to add a threshold (not the optimal one, but a reasonably sub optimal).
There are a lot of discussions on this already. I believe, deviating from pseudocode should be with the concern of @norvig sir.

thevarunsharma:
But the code is similar to the pseudocode provided here [https://github.com/aimacode/aima-pseudocode/blob/master/md/Cross-Validation-Wrapper.md](url)

thevarunsharma:
but the code is similar to the pseudocode at https://github.com/aimacode/aima-pseudocode/blob/master/md/Cross-Validation-Wrapper.md

SiluPanda:
Yes, it is similar. Let's wait for contributors to add their feedback.

dmeoli:
Thanks @thevarunsharma. learning.py has since been refactored and cross_validation_wrapper reimplemented (the old TODO version this patched is gone), so this no longer applies. Closing as superseded — thanks!

PR #1056, opened 2019-03-31
single_layer_Auto_encoder.py

uchihabloodline:
here is the representation by using external libraries(keras) of a single layer Auto encoder.


dmeoli:
Thanks @uchihabloodline. deep_learning4e.py already provides AutoencoderLearner, and the repo doesn't keep standalone Keras scripts at the root, so this single_layer_Auto_encoder.py duplicates existing functionality. Closing as superseded — thanks!

PR #1059, opened 2019-04-02
The function truncated_svd() should not return negative singular values.

mikhail-j:
@Chipe1 @norvig Singular values by definition are non-negative, but currently truncated_svd() returns negative singular values at times.

This pull request modifies truncated_svd() and its respective truncated_svd() tests to always return non-negative real singular values.


norvig:
Good job, thanks!

PR #1065, opened 2019-04-11
added the mentioned AC4 algorithm for constraint propagation

DonatoMeoli:
AC3 algorithm has non-optimal worst-case time-complexity O(cd^3 ), while AC4 algorithm runs in O(cd^2) worst-case time. [Here](http://www.dcs.gla.ac.uk/~pat/cpM/papers/GAC/ac4.pdf) the original paper by Mohr and Henderson.


DonatoMeoli:
Is it possible to add the code of algorithms whose pseudocode is not present in the book but is only mentioned?

Issue #1068, opened 2019-04-12
Improve the python implementation of some search algorithms


codefeeder:
### Issue Type
Code improvement

### Changes
Returning the complete path from the root node to the goal node from existing functions for search algorithms in search.py

### Descriptions
The current recursive and iterative algorithms for breadth-first search, depth-first search, etc. only returns the node or None when goal node is found or not found respectively.
I propose to improve the algorithm to return the complete path from the start node to the goal node in case a path is found otherwise return None.

I also propose to modify driver functions accordingly.

@MrDupin may I work on this?

MrDupin:
This sounds good. What do you mean by 'driver functions' though?

codefeeder:
By 'driver functions' I am referring to the functions calling the search functions as their code also would need to be modified to accommodate  the changes in the return values of the search functions.

MrDupin:
Hmmm, if the driver function changes are only in the way they get return values, then it's fine. I don't want to modify them that much, since they must be as close to the pseudocode as possible.

codefeeder:
@MrDupin I have made the changes and as suggested by you lso, I made minimal changes in the test_search.py file in PR #1070 
My changes pass all the test cases provided in test_search.py but during final testing the output is expected in the old format and therefore, some test cases failed.
Should I format the output according to the old requirements?

MrDupin:
What is the old format and what is the new format of the output?

codefeeder:
![image](https://user-images.githubusercontent.com/22821850/56082119-8c7eb280-5e32-11e9-813c-d249b27f3a8e.png)
![image](https://user-images.githubusercontent.com/22821850/56082128-a6b89080-5e32-11e9-886a-bff67a9b51ee.png)
The old format("<Node ()>") is created by using the single node returned by the search functions.
But, the newly modified search program is directly returning the list of nodes in the path without any further computation.
But due to the difference in the formats, the above test case failed even though the values were same.

MrDupin:
Hmm, you can change the format, sure. What I'm not sure about now is the implementation itself. It seems that you are using quite a lot more lines to find the full path. Can we not use `Node`'s `parent` attribute to backtrack through the path?

codefeeder:
Yes, sir, we can but that would increase the computation by again backtracking from the current node to the root which according to me can be done at the same time we finding a path.

MrDupin:
[...]

PR #1070, opened 2019-04-13
BFS_Tree, DFS_Tree, BFS_Graph, DFS_Graph, Best First Search now returns complete path

codefeeder:
Resolves issue - #1068

This PR is to add the functionality of returning the complete path from the search functions - BFS and DFS for trees, BFS and DFS for graphs and Best First Search.
Details mentioned in issue #1068 

I have followed the contributing guidelines and have provided proper comments.

File Edited - search.py, test_search.py, text.py

### DIFF
--- a/search.py
+++ b/search.py
@@ -184,27 +184,15 @@ def breadth_first_tree_search(problem):
     """Search the shallowest nodes in the search tree first.
         Search through the successors of a problem to find a goal.
         The argument frontier should be an empty queue.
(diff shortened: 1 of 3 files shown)


dmeoli:
Thanks @codefeeder. This changes the core search functions to return a path string instead of the goal Node, which would break everything that relies on the returned node's .solution()/.path(). The path is already available via node.solution()/node.path(). Closing as it changes the core API; a focused, non-breaking helper would be welcome. Thanks!

PR #1074, opened 2019-04-22
SVM is added

ashishgit7:
SVM is added as per chapter 18.9


dmeoli:
Thanks @ashishgit7. learning.py already includes an SVM implementation (SVC/SVM via qpsolvers' solve_qp, plus multi-class classifiers), so this addition is superseded. Closing — thanks for contributing!

PR #1075, opened 2019-06-05
Bruyang

bruceyoungsysu:
- Add Monte Carlo Tree Search and tests on TTT and ConnectFour.
- Change the input of model based reflex agent in chapter 2. Modify its test.


games4e.py
```diff
@@ -0,0 +1,630 @@
...
+            n.U += utility
+        # if utility == 0:
+        #     n.U += 0.5
+        n.N += 1
+        if n.parent:
+            backprop(n.parent, -utility)
+
+    root = MCT_Node(state=state)
+
+    while N > 0:
```
norvig: Why a while loop here? Why not "for _ in range(N)"?

PR #1078, opened 2019-07-08
Update games.py - allow query_player to get user input before returning

SandersLin:
The original code bails when user fail to input valid moves. One consequence of this when the user inputs invalid string (typo ect.) , `query_player ` returns None and this messes up the entire playing sequence. (the other player gets to play two ply in a row)

### DIFF
--- a/games.py
+++ b/games.py
@@ -181,11 +181,14 @@ def query_player(game, state):
     print("")
     move = None
     if game.actions(state):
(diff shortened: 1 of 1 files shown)


dmeoli:
Thanks @SandersLin! Your fork branch was gone, so we rebased this onto master and merged it as the new PR (authorship preserved). Closing in favor of it.

PR #1079, opened 2019-07-08
update games.py connect 4 display method

SandersLin:
original code displays board sideways. Fixed display method to print board bottom down


dmeoli:
Thanks @SandersLin. ConnectFour.display has since been rewritten on master (it now prints per-point W/B counts), so this 2019 display change no longer applies. Closing as superseded — thanks!

PR #1081, opened 2019-07-08
Update games.py

SandersLin:
Since the player function passed in is assumed to only accept (game,state) 2 arguments, I had to write a wrapper around alphabeta_cutoff_search() function.

### DIFF
--- a/games.py
+++ b/games.py
@@ -201,6 +201,8 @@ def alphabeta_player(game, state):
 def expectiminimax_player(game, state):
     return expectiminimax(state, game)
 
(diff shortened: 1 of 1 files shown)


dmeoli:
Thanks @SandersLin. This wraps alphabeta_cutoff_search, but that function was since renamed to alpha_beta_cutoff_search, so the PR no longer applies as-is. The repo also already exposes alpha_beta_player; a small cutoff-player wrapper could be added in a fresh PR if useful. Closing for now — thanks!

PR #1084, opened 2019-07-18
Fixed the error in function best_first_graph_search_for_vis() in search…

qinhuah:
In line **1523**, _f(incumbent)_ is actually do not exist. The same happened in **1524** for _frontier[incumbent]_.
This two bugs will lead to a error of "int object do not have a path-cost attribute" when runing a-star algorithm on romania map problem.


dmeoli:
Thanks @qinhuah. best_first_graph_search_for_vis was fixed in #1242 and runs correctly on current master (verified a search Arad->Bucharest returns the expected path). Note the change proposed here, del incumbent, wouldn't remove the node from the frontier (incumbent is the stored priority value, a number) - master correctly uses del frontier[child]. Closing as superseded by #1242. Thanks!

PR #1090, opened 2019-07-25
Perception

bruceyoungsysu:
- chapter 24 perception and tests
- Faster RCNN model under aima-data


antmarakis:
While your contributions are useful, is it possible to cut down on the number of different commits? Right now you have too many for us to walk through, which makes reviewing the overall PR quite difficult.

PR #1092, opened 2019-07-29
added map coloring SAT problem

DonatoMeoli:


csp.py
```diff
@@ -356,7 +365,7 @@ def build_topological(node, parent, neighbors, visited, stack, parents):
     visited[node] = True
 
     for n in neighbors[node]:
-        if(not visited[n]):
+        if (not visited[n]):
```
antmarakis: I think the parentheses can be removed completely.

DonatoMeoli:
Ok, done! (:

Issue #1098, opened 2019-08-14
probability.ipynb: enumeration_ask and elimination_ask runtime comparison wrong


alessandrocuda:
By running **probability.ipynb** I noticed an error in the result of **Runtime comparison** between `enumeration_ask` and `elimination_ask` algorithms.
In the notebook is reported that:
> We observe that variable elimination was faster than enumeration as we had expected...

but by running the code you get the opposite result:

```
%%timeit -n 1000
enumeration_ask('Burglary', dict(JohnCalls=True, MaryCalls=True), burglary).show_approx()
# 104 µs ± 26.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

```
%%timeit -n 1000
elimination_ask('Burglary', dict(JohnCalls=True, MaryCalls=True), burglary).show_approx()
# 243 µs ± 19.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

ad71:
Hi @alessandrocuda I am responsible for this. When I was working on this, I wrote the results I got, but I can't explain this. Maybe it has something to do with number of threads and how python tries to optimize things, or maybe it is version-specific? I'm sorry I don't know what happened.

alessandrocuda:
Thank you very much for the quick reply @ad71 . I'm investigating why this happens but it could be any one of those listed, at least it might be an idea to update the comparison on the notebook in a formal way? I could open a pull request and update the section with this information

elimination_ask should be a optimization of enumeration_ask but the complexity of some operations requires attention:

- `sum_out` and `pointwise_product` function used in elimination_ask is where space and time complexity arise in the variable elimination algorithm (AIMA3e pg. 526) (look at the profiler result [profiler_result.txt](https://github.com/aimacode/aima-python/files/3502587/profiler_result.txt))

- Elimination ordering is important, in this case the algorithm apply a reversed order (AIMA3e pg. 527)

-  An optimization (AIMA3e pg. 528) is to remove 'every variable that is not an ancestor of a query variable or evidence variable as it is irrelevant to the query' (i.e. sums to 1). 

- The subset of variables from the Bayesian Network to be retained after irrelevant hidden variables have been removed.

ad71:
Sure, that will be very helpful. Thanks!

PR #1100, opened 2019-08-15
Update probability.ipynb: fixed issue #1098

alessandrocuda:
fixed issue  https://github.com/aimacode/aima-python/issues/1098


probability.ipynb
```diff
@@ -74,12 +72,21 @@
       "text/html": [
        "<!DOCTYPE html PUBLIC \"-//W3C//DTD HTML 4.01//EN\"\n",
        "   \"http://www.w3.org/TR/html4/strict.dtd\">\n",
-       "\n",
+       "<!--\n",
```
ad71: Can we remove this comment? Are we actually using Pygments?

probability.ipynb
```diff
@@ -74,12 +72,21 @@
...
+       "generated by Pygments <http://pygments.org>\n",
+       "Copyright 2006-2019 by the Pygments team.\n",
+       "Licensed under the BSD license, see LICENSE for details.\n",
+       "-->\n",
        "<html>\n",
        "<head>\n",
        "  <title></title>\n",
        "  <meta http-equiv=\"content-type\" content=\"text/html; charset=None\">\n",
        "  <style type=\"text/css\">\n",
+       "/*\n",
```
ad71: Same here, and all the ones below. If we are not using Pygments, it'd be better to not have these comments here. A simple find-replace would do the job.

alessandrocuda: Thanks! removed all the irrelevant lines and made a new commit. It should be all right now.

probability.ipynb
```diff
@@ -3015,10 +3014,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We observe that variable elimination was faster than enumeration as we had expected but the gain in speed is not a lot, in fact it is just about 30% faster.\n",
+    "In this test case we observe that variable elimination it's slower than what we expected. It has something to do with number of threads, how python tries to optimize things and  this happens because the network is very small, with just 5 nodes. The `elimination_ask` has some critical point and some optimizations must be perfomed as seen above.\n",
```
antmarakis: It should be 'is slower' instead of 'it's slower'. Also, it's best if you capitalize 'Python'.

probability.ipynb
```diff
@@ -2966,6 +2937,34 @@
     "elimination_ask('Burglary', dict(JohnCalls=True, MaryCalls=True), burglary).show_approx()"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Elimination ask optimizations\n",
```
antmarakis: This should be in title form, with every other word capitalized: Elimination Ask Optimization

probability.ipynb
```diff
@@ -2966,6 +2937,34 @@
...
+    "\n",
+    "  `sum_out` and `pointwise_product` function used in `elimination_ask` is where space and time complexity arise in the variable elimination algorithm (AIMA3e pg. 526).\n",
+    "\n",
+    ">The only trick is to notice that any factor that does not depend on the variable to be summed out can be moved outside the summation.\n",
+    "\n",
+    "- **Variable ordering**:\n",
+    "\n",
+    "  Elimination ordering is important, every choice of ordering yields a valid algorithm, but different orderings cause different intermediate factors to be generated during the calculation (AIMA3e pg. 527). In this case the algorithm apply a reversed order.\n",
+    "\n",
+    "> In general, the time and space requirements of variable elimination are dominated by\n",
```
antmarakis: Here I see a newline between 'by' and 'the size'.

probability.ipynb
```diff
@@ -2966,6 +2937,34 @@
...
+    "\n",
+    "- **Operation on factors**:\n",
+    "\n",
+    "  `sum_out` and `pointwise_product` function used in `elimination_ask` is where space and time complexity arise in the variable elimination algorithm (AIMA3e pg. 526).\n",
+    "\n",
+    ">The only trick is to notice that any factor that does not depend on the variable to be summed out can be moved outside the summation.\n",
+    "\n",
+    "- **Variable ordering**:\n",
+    "\n",
+    "  Elimination ordering is important, every choice of ordering yields a valid algorithm, but different orderings cause different intermediate factors to be generated during the calculation (AIMA3e pg. 527). In this case the algorithm apply a reversed order.\n",
```
antmarakis: "In this case the algorithm apply a reversed order" Here, is 'algorithm':

a) plural: "In this case the algorithms apply a reversed order"

b) singular: "In this case the algorithm applies a reversed order"

probability.ipynb
```diff
@@ -3015,10 +3014,9 @@
...
+    "In this test case we observe that variable elimination it's slower than what we expected. It has something to do with number of threads, how python tries to optimize things and  this happens because the network is very small, with just 5 nodes. The `elimination_ask` has some critical point and some optimizations must be perfomed as seen above.\n",
     "<br>\n",
-    "This happened because the bayesian network in question is pretty small, with just 5 nodes, some of which aren't even required in the inference process.\n",
-    "For more complicated networks, variable elimination will be significantly faster and runtime will reduce not just by a constant factor, but by a polynomial factor proportional to the number of nodes, due to the reduction in repeated calculations."
+    "Of course, for more complicated networks, variable elimination will be significantly faster and runtime will reduce not just by a constant factor, but by a polynomial factor proportional to the number of nodes, due to the reduction in repeated calculations."
```
antmarakis: It is "will drop not just" instead of "will reduce not just".

Issue #1053, opened 2019-03-26
Unification algorithm


msimi:
>>> from agents import *
>>> from logic import *
>>> A, B, C, D, E, F, G, P, Q, x, y, z, F, G, u = map(Expr, 'ABCDEFGPQxyzFGu')
>>> unify(P(A, x, F(G(y))), P(z, F(z), F(u)))
{z: A, x: F(z), u: G(y)}
=> I think this should return instead {z: A, x: F(A), u: G(y)}

The unifier returned does not make the two expressions identical as it should
>>> subst({z: A, x: F(z), u: G(y)}, P(A, x, F(G(y))))
P(A, F(z), F(G(y)))
>>> subst({z: A, x: F(z), u: G(y)}, P(z, F(z), F(u)))
P(A, F(A), F(G(y)))

SiluPanda:
Ma'am, can you kindly post the code file name? Or the link to this code snippet?

msimi:
The file is logic.py from the AIMA-Python repository and I am referring to the function unify within it
Implementing unification.

— M. Simi

dmeoli:
Unification is implemented in `logic.py`: `unify` (Robinson's algorithm) plus `unify_var`, `occur_check`, `subst`, and a second Martelli–Montanari variant `unify_mm`. The specific non-normal-form result reported here is also fixed — `cascade_substitution` (called from `unify_var`) normalizes the substitution, so e.g. `unify(P(A, x, F(G(y))), P(z, F(z), F(u)))` now returns `{z: A, x: F(A), u: G(y)}`, and substituting it into both expressions yields the identical `P(A, F(A), F(G(y)))`. Closing as implemented.

PR #1101, opened 2019-08-16
Fix for unify algorithm in logic.py

alessandrocuda:
These commits should resolve issue (#1053)
It seems that this bug is well known in the [aima-java repo](https://github.com/aimacode/aima-java/blob/AIMA3e/aima-core/src/main/java/aima/core/logic/fol/Unifier.java) but was not corrected in the aima-python repo. In the `unify_var(var, x, s)` method a **cascade_substitution** must be perform after adding a new mapping var/x to the substitution set 's' as can be seen from the unification algorithm developed in java in the [aima-java repo](https://github.com/aimacode/aima-java/blob/AIMA3e/aima-core/src/main/java/aima/core/logic/fol/Unifier.java).

### DIFF
--- a/logic.py
+++ b/logic.py
@@ -1417,13 +1417,13 @@ def subst(s, x):
         return Expr(x.op, *[subst(s, arg) for arg in x.args])
 
 def cascade_substitution(s):
-    '''This method allow to return a correct unifier in normal form
+    """This method allows to return a correct unifier in normal form
     and perform a cascade substitution to s.
     For every mapping in s perform a cascade substitution on s.get(x)
     and if it is replaced with a function ensure that all the function 
-    terms are correctly updates by passing over them again.
+    terms are correct updates by passing over them again.
 
-    This fix issue: https://github.com/aimacode/aima-python/issues/1053
+    This issue fix: https://github.com/aimacode/aima-python/issues/1053
     unify(expr('P(A, x, F(G(y)))'), expr('P(z, F(z), F(u))')) 
     must return {z: A, x: F(A), u: G(y)} and not {z: A, x: F(z), u: G(y)}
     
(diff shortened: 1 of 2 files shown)


logic.py
```diff
@@ -1415,6 +1417,33 @@ def subst(s, x):
     else:
         return Expr(x.op, *[subst(s, arg) for arg in x.args])
 
+def cascade_substitution(s):
+    '''This method allow to return a correct unifier in normal form
```
antmarakis: In the repo we generally use double quotes for docstrings. I would prefer it that way for consistency's sake.

Also, it should be 'allows' instead of 'allow'.

logic.py
```diff
@@ -1415,6 +1417,33 @@ def subst(s, x):
...
+    and perform a cascade substitution to s.
+    For every mapping in s perform a cascade substitution on s.get(x)
+    and if it is replaced with a function ensure that all the function 
+    terms are correctly updates by passing over them again.
```
antmarakis: It should either be 'correct updates' or 'correctly updated'.

logic.py
```diff
@@ -1415,6 +1417,33 @@ def subst(s, x):
...
+
+    This fix issue: https://github.com/aimacode/aima-python/issues/1053
```
antmarakis: 'This issue fix' or 'This fixes the issue' (preferably the former).

logic.py
```diff
@@ -1415,6 +1417,33 @@ def subst(s, x):
...
+    {x: G(z), y: G(z)}
+    
+    Parameters
+    ----------
+    s : Dictionary
+        This contain a substution'''
+
+    for x in s:
+        s[x] = subst(s, s.get(x))
+        if isinstance(s.get(x),Expr) and not is_variable(s.get(x)):
```
antmarakis: There is a space missing in `isinstance`, between the arguments.

logic.py
```diff
@@ -1415,6 +1417,33 @@ def subst(s, x):
...
+        # Ensure Function Terms are correctly updates by passing over
```
antmarakis: It should either be 'correct updates' or 'correctly updated'.

logic.py
alessandrocuda: I reviewed all the requests and committed the relative commit to fix them but I don't understand why The Travis CI build failed

alessandrocuda: All checks have passed in the new commit! It should be all right now.

PR #1102, opened 2019-08-17
Demo of chapter 19 (4th eidtion)

bruceyoungsysu:
- learners
- Loss functions and layers
- Optimizer and backpropagation
- RNN


delaray:
Very good @bruceyoungsysu !

PR #1106, opened 2019-08-22
Demo notebooks of Chapter 16 for 4th edition

bruceyoungsysu:
- introduction
- MDPs
- Algorithms for MPDs
- Partially Observed MDPs
- Sequential Decision Problem
- MDP apps
- chapter 17 should be put into the part of planning


dmeoli:
Thanks a lot for these, @bruceyoungsysu. We've landed five of the six notebooks (Introduction, MDPs, Algorithms for MDPs, Sequential Decision Problems, Partially Observable MDP) under notebooks/chapter16-17/ in #1329, with your authorship preserved — they fill the 4e notebook gap for MDPs and run against the current code. We left out mdp_apps.ipynb because it targets stale 3e APIs that no longer exist on master (from mdp/notebook import *, from utils import argmax, plot_utility); it would need a separate rewrite. Closing in favor of #1329 — thanks for the great contribution!

PR #1108, opened 2019-08-23
Making simple decision (chapter 15 and 25)

bruceyoungsysu:
- Chapter 15 making simple decisions and chapter 25 Monte Carlo localization


delaray:
Nice!

PR #1113, opened 2019-09-12
fixed expanded_actions( ), added CSPlan with n-ary CSP definition, problems and tests, AC3b and AC4 with tests

### LINKED ISSUE #1111: planning.py: expanded_actions( ) generates actions which doesn't satisfy the arguments' domain
For example, for the three_block_tower problem:

`from planning import *`
`three_block_tower().expand_actions()`

should returns: 

`[Move(B, A, C), Move(B, C, A), Move(B, Table, A), Move(B, Table, C), Move(A, B, C), Move(A, C, B), Move(A, Table, B), Move(A, Table, C), Move(C, B, A), Move(C, A, B), Move(C, Table, B), Move(C, Table, A), MoveToTable(B, A), MoveToTable(B, C), MoveToTable(A, B), MoveToTable(A, C), MoveToTable(C, B), MoveToTable(C, A)]`

but it also returns actions such as `Move(Table, _, _)`, `Move(_, _, Table)`, `MoveToTable(Table, _)` or `MoveToTable(_, Table)` which are illegal because doesn't satisfy the arguments' domain.

DonatoMeoli:
- fixed #1111 
- added the CSPlan [Section 10.4.3] with the definition of a CSP with n-ary constraints and some problems including the cryptarithmetic puzzle T W O + T W O = F O U R [Figure 6.2]
- AC3b, an improved version of AC-3 with double-support check domain-heuristic which has the same complexity of AC-3, O(cd^3), but is faster on average cases
- AC4 (page 210, footnote 1) which runs in O(cd^2) worst-case time but can be slower than AC-3 on average cases


csp.py
```diff
@@ -744,3 +747,526 @@ def solve_zebra(algorithm=min_conflicts, **args):
...
+    @staticmethod
+    def is_word(words):
+        """Returns True if the letters concatenated form a word in words, False otherwise"""
+
+        def isw(*letters):
+            return "".join(letters) in words
+
+        return isw
+
+    @staticmethod
```
antmarakis: Is it important to mark these as `staticmethod`? I don't think we use it anywhere else in the project and I would prefer to keep things simple.

DonatoMeoli:
All these methods labeled as static are possible common usage constraints
(like allDiff) that a Constraint object can take as a second parameter. If
you prefer, I can move them outside the class as simple functions.

PR #1116, opened 2019-09-20
fixing names

josd:
running csp.ipynb was giving NameError for australia, usa, france
correcting to australia_csp, usa_csp, france_csp works fine


DonatoMeoli:
Thx Josd! It was my mistake: after adding the problems also as SAT I renamed those as CSP as "ausralia_csp", "france_csp", etc. and those as SAT as "australia_sat", "france_sat", etc. but I would have to check also the .ipynb! Sorry. :)

PR #1120, opened 2019-10-03
moved util functions to utils.py, moved probability learners from learning.py to probabilistic_learning.py with tests, fixed typos and fixed imports in .ipynb files

DonatoMeoli:


antmarakis:
There are some minor conflicts, if you could fix those it would be great. Also, I am going to wait a couple of days for Travis to stop malfunctioning.

PR #1130, opened 2019-11-02
added csp, logic, planning and probability .ipynb

DonatoMeoli:


antmarakis:
I am a bit confused, what is the purpose of this PR? It seems that (mostly) the aforementioned notebooks are broken up into smaller ones. Do I understand correctly? Viewing all these files individually is going to be time-consuming.

DonatoMeoli:
This PR aims to add many other details and comparing, for example, about all the arc consistency algorithms available in csp.py (AC3, AC4, AC3b, etc.) giving details about the performance of each of them or, again, about all the classical planning approaches (as SAT, as CSP, as forward or backward search, etc.) giving, even here, details about the performance. 
It's much more than basic files, there is no redundancy!

PR #1135, opened 2019-11-19
added binary and multiclass SVM with tests

DonatoMeoli:


learning.py
```diff
@@ -810,7 +815,110 @@ def init_examples(examples, idx_i, idx_t, o_units):
...
+        self.sv_x, self.sv_y, self.alphas = X[sv_indices], y[sv_indices], self.alphas[sv_indices]
+        self.n_sv = len(sv_indices)
+        if self.kernel == linear_kernel:
+            self.w = np.dot(self.alphas * self.sv_y, self.sv_x)
+        # calculate b: average over all support vectors
+        sv_boundary = self.alphas < self.C - self.eps
+        self.b = np.mean(self.sv_y[sv_boundary] - np.dot(self.alphas * self.sv_y,
+                                                         self.kernel(self.sv_x, self.sv_x[sv_boundary])))
+
+    def QP(self, X, y):
```
antmarakis: In this function, isn't it possible to rename the variables to something more descriptive? I am not familiar with SVM literature, but I think it would be best if the variable names described a bit more what each variable does.

learning.py
```diff
@@ -810,7 +815,110 @@ def init_examples(examples, idx_i, idx_t, o_units):
...
+        q = -np.ones(m)
+        G = np.vstack((-np.identity(m), np.identity(m)))
+        h = np.hstack((np.zeros(m), np.ones(m) * self.C))
+        A = y.reshape((1, -1))
+        b = np.zeros(1)
+        # make sure P is positive definite
+        P += np.eye(P.shape[0]).__mul__(1e-3)
+        self.alphas = solve_qp(P, q, G, h, A, b, sym_proj=True)
+
+    def project(self, x):
```
antmarakis: I would like to see a comment about what this function does.

learning.py
```diff
@@ -810,7 +815,110 @@ def init_examples(examples, idx_i, idx_t, o_units):
...
+        self.kernel = kernel
+        self.decision_function = decision_function
+        self.C = C
+        self.n_class, self.classifiers = 0, []
+
+    def fit(self, X, y):
+        labels = np.unique(y)
+        self.n_class = len(labels)
+        if self.decision_function == 'ovr':  # one-vs-rest method
+            for label in labels:
```
antmarakis: Again here, the variable names could be a bit more descriptive.

learning.py
```diff
@@ -810,7 +815,110 @@ def init_examples(examples, idx_i, idx_t, o_units):
...
+                    x1, y1 = np.r_[X[neg_id], X[pos_id]], np.r_[y[neg_id], y[pos_id]]
+                    y1[y1 == labels[i]] = -1.0
+                    y1[y1 == labels[j]] = 1.0
+                    clf = BinarySVM(self.kernel, self.C)
+                    clf.fit(x1, y1)
+                    self.classifiers.append(copy.deepcopy(clf))
+        else:
+            return ValueError("Decision function must be either 'ovr' or 'ovo'.")
+
+    def predict(self, X):
```
antmarakis: It would be great if there was some sort of description of what the code here does.

reinforcement_learning4e.py
```diff
@@ -310,7 +315,7 @@ def __call__(self, percept):
             self.s = self.a = self.r = None
         else:
             self.s, self.r = s1, r1
-            self.a = argmax(actions_in_state(s1), key=lambda a1: self.f(Q[s1, a1], Nsa[s1, a1]))
+            self.a = max(actions_in_state(s1), key=lambda a1: self.f(Q[s1, a1], Nsa[s1, a1]))
```
antmarakis: I have been seeing you replace argmax with max a lot and I want to make sure the replacement is correct, because typically argmax and max return different things (index vs value).

mdp4e.py
```diff
@@ -517,38 +514,3 @@ def pomdp_value_iteration(pomdp, epsilon=0.1):
                 },
     }
 """
-
-
-# __________________________________________________________________________
-# Chapter 17 Multiagent Planning
-
-
-def double_tennis_problem():
```
antmarakis: Where did this function go?

tests/test_utils.py
```diff
@@ -163,62 +151,17 @@ def test_dot_product():
     assert dot_product([1, 2, 3], [0, 0, 0]) == 0
 
 
-def test_element_wise_product():
```
antmarakis: Where did these functions go?

utils.py
```diff
@@ -147,74 +150,35 @@ def histogram(values, mode=0, bin_function=None):
...
     return tuple(map(operator.add, a, b))
 
 
-def scalar_vector_product(X, Y):
+def scalar_vector_product(x, y):
     """Return vector as a product of a scalar and a vector"""
-    return [X * y for y in Y]
-
-
-def scalar_matrix_product(X, Y):
```
antmarakis: Same here, where are these functions now?

utils4e.py
```diff
@@ -277,33 +247,17 @@ def vector_add(a, b):
...
+    """Return vector as a product of a scalar and a vector recursively."""
+    return [scalar_vector_product(x, _y) for _y in y] if hasattr(y, '__iter__') else x * y
 
 
-def scalar_matrix_product(X, Y):
-    """Return matrix as a product of a scalar and a matrix"""
-    return [scalar_vector_product(X, y) for y in Y]
-
-
-def inverse_matrix(X):
```
antmarakis: Same here again, what replaces these functions?

reinforcement_learning4e.py
DonatoMeoli: Exactly! "np.argmin" and "np.argmax" returns the position of the minimum and maximum value respectively, while the builtins "min" and "max" just the minimum and maximum value respectively.
Bit the different usages of these two functions seem was not clear to anyone because because the role of "min" and "max" was played by "np.argmin" and "np.argmax", in fact in utils.py there was a line of code "argmin = min" and "argmax = max". I've just fixed all the occurrences of argmin and argmax replacing these with just min or max and with "np.argmin" and "np.argmax" when necessary (for example in Viterbi algorithm, learning.py, both were useful).

mdp4e.py
DonatoMeoli: This planning problem is already defined in planning.py. It was a duplicate code!

tests/test_utils.py
DonatoMeoli: This is an inefficient way to implement operations between matrices, I think that the focus of the book is the intelligent algorithms so operations between matrices of this type can be delegated to numpy that performs them in a more efficient manner.

utils.py
DonatoMeoli: Same here.

utils4e.py
DonatoMeoli: Again, the same here. I replaced it with numpy.linalg.inv.

learning.py
DonatoMeoli: The name of the variables is common in the SVM scope, anyway I added doc.

learning.py
DonatoMeoli: Renamed to make it more intuitive and added doc.

learning.py
DonatoMeoli: Same here.

PR #1142, opened 2019-12-24
MAINT: Add documentation and descriptive variable names in search.py

tirthasheshpatel:
Add:
 - Docstring to `__hash__` method of `Node` class.
 - Add descriptive names in Bidirectional search that are consistent with the paper.

### DIFF
--- a/search.py
+++ b/search.py
@@ -123,6 +123,10 @@ class Node:
         return isinstance(other, Node) and self.state == other.state
 
     def __hash__(self):
+        # We use the hash value of the state
+        # stored in the node instead of the node
+        # object itself to quickly search a node
+        # with the same state in a Hash Table
         return hash(self.state)
 
 
--- a/search.py
+++ b/search.py
@@ -340,18 +340,18 @@ def bidirectional_search(problem):
         open_dir.remove(n)
         closed_dir.append(n)
 
-        for child in problem.actions(n):
-            if child in open_dir or child in closed_dir:
-                if g_dir[child] <= problem.path_cost(g_dir[n], n, None, child):
+        for c in problem.actions(n):
+            if c in open_dir or c in closed_dir:
+                if g_dir[c] <= problem.path_cost(g_dir[n], n, None, c):
                     continue
 
-                open_dir.remove(child)
+                open_dir.remove(c)
 
-            g_dir[child] = problem.path_cost(g_dir[n], n, None, child)
-            open_dir.append(child)
+            g_dir[c] = problem.path_cost(g_dir[n], n, None, c)
+            open_dir.append(c)
 
-            if child in open_other:
-                U = min(U, g_dir[child] + g_other[child])
+            if c in open_other:
+                U = min(U, g_dir[c] + g_other[c])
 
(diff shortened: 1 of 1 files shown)


search.py
```diff
@@ -336,42 +340,44 @@ def extend(U, open_dir, open_other, g_dir, g_other, closed_dir):
         open_dir.remove(n)
         closed_dir.append(n)
 
-        for c in problem.actions(n):
-            if c in open_dir or c in closed_dir:
-                if g_dir[c] <= problem.path_cost(g_dir[n], n, None, c):
+        for child in problem.actions(n):
```
tirthasheshpatel: I am very confused here! Why is the child present in `problem.actions(n)` instead of `n.expand(problem)`? 
Why does `problem.action(n)` return the next states instead of actions possible in that states?

tirthasheshpatel: Inspecting further, it turns out that bidirectional search only supports `GraphProblem` type problems. I think we can add support for all the problem types by storing the nodes instead of states in open/closed list for both direction and calling n.expand(problem) on it to return the next states. Any other ideas?

antmarakis: `problem.actions(n)` does return the actions possible in `n`, but since this is a graph problem, the actions possible are the next states. The function simply returns the neighbouring nodes, which are both the next states and actions possible. I believe though that you are right, if we want to support other types of problems, we need to switch to `n.expand(problem)` which is more general.

If you want, you can take this up, making bidirectional search general. That sounds like a great contribution!

antmarakis: I personally feel like `c` is descriptive enough, since in a lot of code c is used to denote the child.

search.py
```diff
@@ -336,42 +340,44 @@ def extend(U, open_dir, open_other, g_dir, g_other, closed_dir):
...
+                U = min(U, g_dir[child] + g_other[child])
 
         return U, open_dir, closed_dir, g_dir
 
     def find_min(open_dir, g):
         """Finds minimum priority, g and f values in open_dir"""
-        m, m_f = np.inf, np.inf
+        # pr_min_f isn't forward pr_min instead it's the f-value
+        # of node with priority pr_min.
+        pr_min, pr_min_f = np.inf, np.inf
```
antmarakis: I agree, the new variable names are indeed more descriptive, nice catch!

search.py
```diff
@@ -336,42 +340,44 @@ def extend(U, open_dir, open_other, g_dir, g_other, closed_dir):
...
+            pr_min_f = min(pr_min_f, f)
 
-        return m, m_f, min(g.values())
+        return pr_min, pr_min_f, min(g.values())
 
     def find_key(pr_min, open_dir, g):
         """Finds key in open_dir with value equal to pr_min
         and minimum g value."""
-        m = np.inf
+        gmin = np.inf
```
antmarakis: Here again I think that `m` is fine, since in what this does is a simple min search. In such code, with only one min, just writing `m` should be enough.

search.py
tirthasheshpatel: Thanks!

search.py
tirthasheshpatel: Yup, will fix that!

PR #1143, opened 2019-12-26
cd into aima folder before installing requirements

Antzelino:


Antzelino:
another alternative could be to cd into the folder right after `git clone`

Antzelino:
any idea why there is a failed check for changing something on the readme?

PR #1147, opened 2020-01-06
[WIP] ENH: add support for all types of problems in Bidirectional Search

### LINKED ISSUE #1144: Add support for all types of problems in Bidirectional Search
Bidirectional search currently only supports `GraphProblem` type problems. I think we can add support for all the problem types by storing the nodes instead of states in open/closed list for both directions and calling `n.expand(problem)` on it to return the next states.

tirthasheshpatel:
Fixes #1144 

I have added support for all types of problems in Bidirectional Search and their tests.

I am still confused about two things:
  1. How to return the node containing the solution path along with the `U` value in BS? Maybe, we can also add that like all other search algorithms.
  2. We need to traverse through each node to calculate the `e` (minimum edge value) value which can take exponential time for large problems. So, is it fine to initialize it to 0 or there is a way to calculate it?

### DIFF
--- a/search.py
+++ b/search.py
@@ -329,7 +329,7 @@ def iterative_deepening_search(problem):
 def bidirectional_search(problem):
     e = problem.find_min_edge()
     gF, gB = {problem.initial: 0}, {problem.goal: 0}
-    openF, openB = [problem.initial], [problem.goal]
+    openF, openB = [Node(problem.initial)], [Node(problem.goal)]
     closedF, closedB = [], []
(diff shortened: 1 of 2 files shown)


antmarakis:
1. Implementing this functionality like in the rest of the algorithms sounds like the way to go, definitely!

2. I think if we initialize them at 0 it would be fine. Maybe add a TODO comment next to the initialization to let people know maybe a better initialization can be explored.

PR #1148, opened 2020-01-07
fixed grabbing behaviour in agent

omar-3:
When a thing is grabbed, it would be hidden from the environment. When released or the agent is removed it would be added to environment again at the same location of the agent. For the issue of grabbed thing movement, it is already implemented. No test has failed.

Have I missed something?

### DIFF
--- a/agents.py
+++ b/agents.py
@@ -510,14 +510,16 @@ class XYEnvironment(Environment):
             agent.direction += Direction.L
         elif action == 'Forward':
             agent.bump = self.move_to(agent, agent.direction.move_forward(agent.location))
-        #         elif action == 'Grab':
-        #             things = [thing for thing in self.list_things_at(agent.location)
-        #                     if agent.can_grab(thing)]
-        #             if things:
-        #                 agent.holding.append(things[0])
+        elif action == 'Grab':
+            things = [thing for thing in self.list_things_at(agent.location)]
+            if agent.can_grab(things[0]):
+                if things:
+                    agent.holding.append(things[0])
+                    self.delete_thing(things[0])
         elif action == 'Release':
             if agent.holding:
-                agent.holding.pop()
+                dropped = agent.holding.pop()
+                self.add_thing(dropped, location=agent.location)
 
     def default_location(self, thing):
         location = self.random_location_inbounds()
@@ -569,10 +571,11 @@ class XYEnvironment(Environment):
     def delete_thing(self, thing):
         """Deletes thing, and everything it is holding (if thing is an agent)"""
         if isinstance(thing, Agent):
-            for obj in thing.holding:
-                super().delete_thing(obj)
-                for obs in self.observers:
-                    obs.thing_deleted(obj)
+            # for obj in thing.holding:
+            #     super().delete_thing(obj)
+            #     for obs in self.observers:
+            #         obs.thing_deleted(obj)
+            del thing.holding
 
         super().delete_thing(thing)
         for obs in self.observers:
--- a/agents.py
+++ b/agents.py
@@ -959,9 +959,9 @@ class WumpusEnvironment(XYEnvironment):
 
         if isinstance(agent, Explorer) and self.in_danger(agent):
             return
-
+            
         agent.bump = False
-        if action in ['TurnRight', 'TurnLeft', 'Forward','Grab']:
+        if action in ['TurnRight', 'TurnLeft', 'Forward', 'Grab']:
             super().execute_action(agent,action)
             agent.performance -= 1
         elif action == 'Climb':
--- a/agents.py
+++ b/agents.py
@@ -511,14 +511,15 @@ class XYEnvironment(Environment):
         elif action == 'Forward':
             agent.bump = self.move_to(agent, agent.direction.move_forward(agent.location))
         elif action == 'Grab':
-            things = [thing for thing in self.list_things_at(agent.location)]
-            if agent.can_grab(things[0]):
-                if things:
-                    agent.holding.append(things[0])
-                    self.delete_thing(things[0])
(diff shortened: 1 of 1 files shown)


agents.py
```diff
@@ -510,14 +510,16 @@ def execute_action(self, agent, action):
...
             agent.bump = self.move_to(agent, agent.direction.move_forward(agent.location))
-        #         elif action == 'Grab':
-        #             things = [thing for thing in self.list_things_at(agent.location)
-        #                     if agent.can_grab(thing)]
-        #             if things:
-        #                 agent.holding.append(things[0])
+        elif action == 'Grab':
+            things = [thing for thing in self.list_things_at(agent.location)]
+            if agent.can_grab(things[0]):
+                if things:
```
tirthasheshpatel: This line `if things` should be on the top otherwise `things[0]` (accessed in the line above) would raise an exception!

agents.py
```diff
@@ -569,10 +571,11 @@ def random_location_inbounds(self, exclude=None):
...
         """Deletes thing, and everything it is holding (if thing is an agent)"""
         if isinstance(thing, Agent):
-            for obj in thing.holding:
-                super().delete_thing(obj)
-                for obs in self.observers:
-                    obs.thing_deleted(obj)
+            # for obj in thing.holding:
+            #     super().delete_thing(obj)
+            #     for obs in self.observers:
+            #         obs.thing_deleted(obj)
```
tirthasheshpatel: I don't see a reason to comment these lines out. Even though the second line is redundant, the observers of the environment still need to be informed about each thing being deleted.

tirthasheshpatel:
I think this will also need some tests to be added

Okhaledzaki:
yess I will do it in a couple of days and make a pull request

agents.py
Okhaledzaki: yes that is true

agents.py
Okhaledzaki: if we delete all of the agent holdings with `del thing.holding`, and thanks to python memory management. All instants of the agent's holding would vanish from memory even in the observer list. Tell me what you think.

Okhaledzaki:
> I think this will also need some tests to be added

Ahh I realized that after making the pull request, the tests would be easy I guess.

Okhaledzaki:
could anyone help me understand why the checks have failed? If fails for python3.4 only

tirthasheshpatel:
> could anyone help me understand why the checks have failed? It fails for python3.4 only

It has nothing to do with the pr. Just close and reopen and the tests will hopefully pass...

agents.py
tirthasheshpatel: The observers of the environment need to be updated about all the things getting deleted along with the agent. I have not seen `observers` used anywhere, so it depends upon the maintainers of the project to keep or exclude it!

tirthasheshpatel:
Turns out keras has removed support for python 3.4. This can be a problem as much of deep learning still depends on the library.
Ping @antmarakis do you think we also need to remove the support for python 3.4?

tirthasheshpatel:
Here's the upstream issue on keras. [gh-13674](https://github.com/keras-team/keras/issues/13674)

Okhaledzaki:
I guess it is better if I open an issue.

tirthasheshpatel:
@Okhaledzaki I don't think this is our issue but I will leave it upto the maintainers to decide!

Okhaledzaki:
@antmarakis hello, would you please check this PR.

agents.py
```diff
@@ -967,21 +967,16 @@ def execute_action(self, agent, action):
 
         agent.bump = False
         if action == 'TurnRight':
```
antmarakis: I think the if statement could be streamlined if we rewrote it as:

```
if action in ['TurnRight', 'TurnLeft', 'Forward']:
            super().execute_action(agent,action)
            agent.performance -= 1
elif action == 'Grab':
            super().execute_action(agent,action)
```

I think the above looks a bit better.

omar659: I had that in mind, but I thought to myself that being more articulate and reasonably redundant in code would help the student retaining the information **that is my case**.

omar659: But actually seeing it, it really didn't help much.

omar659: We can even add 'Grab' action to the list because the performance also decreases by one in the grabbing action, I guess you didn't notice the line because my code isn't formatted very well.

agents.py
```diff
@@ -966,16 +961,7 @@ def execute_action(self, agent, action):
...
-            super().execute_action(agent,action)
-            agent.performance -= 1
-        elif action == 'TurnLeft':
-            super().execute_action(agent,action)
-            agent.performance -= 1
-        elif action == 'Forward':
-            super().execute_action(agent,action)
-            agent.performance -= 1
-        elif action == 'Grab':
+        if action in ['TurnRight', 'TurnLeft', 'Forward','Grab']:
```
antmarakis: A space is needed after the last comma.

omar659: done

omar659:
something has been broken in the deeplearning module @antmarakis :"D

omar659:
@antmarakis The tests have all passed finally *phew*

agents.py
```diff
@@ -964,24 +959,10 @@ def execute_action(self, agent, action):
...
-            agent.performance -= 1
-        elif action == 'Grab':
-            things = [thing for thing in self.list_things_at(agent.location)
-                      if agent.can_grab(thing)]
-            if len(things):
-                print("Grabbing", things[0].__class__.__name__)
-                if len(things):
-                    agent.holding.append(things[0])
+        if action in ['TurnRight', 'TurnLeft', 'Forward', 'Grab']:
+            super().execute_action(agent,action)
```
antmarakis: You need a space after the comma, to comply with the overall project style.

omar659: done

PR #1150, opened 2020-01-08
fixed typos in gui folder

DonatoMeoli:


Okhaledzaki:
same issue as with pr #1148

antmarakis:
I would like to know, what are you planning on doing with the DL functions you removed? Are you going to add them to `learning`, or something else? I think they should be put in the DL script.

DonatoMeoli:
> I would like to know, what are you planning on doing with the DL functions you removed? Are you going to add them to `learning`, or something else? I think they should be put in the DL script.

I removed the classes that use Keras because I think that the repository's focus is to propose simple and intuitive implementations to the problems proposed in the book and not to wrap other high-level DL libraries.
I hope someone wants to try in the future with the implementation of these classes, maybe I will try in a short future in the context of the Computational Neuroscience exam.

antmarakis:
Thing is, I was not the one who merged these commits, so I am not sure what to do with them. I remember a while ago Dr. Norvig wanted to start using TensorFlow in here, so maybe this is it.

I would prefer if they remained in there, just to be safe.

DonatoMeoli:
> Thing is, I was not the one who merged these commits, so I am not sure what to do with them. I remember a while ago Dr. Norvig wanted to start using TensorFlow in here, so maybe this is it.
> 
> I would prefer if they remained in there, just to be safe.

Ok @antmarakis, in the meanwhile I provided to re-added Keras class in order to give you the possibility to merge the others changes.

PR #1157, opened 2020-01-16
[MRG] ENH: Add animations for search algorithms

### LINKED ISSUE #1155: Visualizations for search algorithms
I browsed through the search algorithms notebook and I think that adding the visualization of how the search algorithm expands and finds the path would be useful and interesting. Kind of what @redblobgames did in his article (and has probably been done in aima-javascript). 

I made this for BFS using ipythonblocks:
vis

tirthasheshpatel:
Fixes #1155 

Add more interactive visualization of search algorithms. I have added almost all the search algorithms but still have left bidirectional search, iterative deepening search, and lrta*.

vis


tirthasheshpatel:
I also did some cleanup in the ``search4e`` notebook!

notebook4e.py
```diff
@@ -1156,3 +1159,251 @@ def plot_pomdp_utility(utility):
...
+        result |= line(x, y, dx, dy, random.choice(lengths))
+    return result
+
+
+def line(x, y, dx, dy, length):
+    """A line of `length` cells starting at (x, y) and going in (dx, dy) direction."""
+    return {(x + i * dx, y + i * dy) for i in range(length)}
+
+
+class AnimateProblem(GridProblem):
```
antmarakis: Isn't there a way to get just the `AnimateProblem`  function in `notebook.py`? I think the rest do not fit here, except `random_lines` and `line`.

tirthasheshpatel: We need methods from the class `GridProblem` (which in turn is a subclass of `Problem`) and removing the class will result in the addition of the same methods in `AnimateProblem` class. I know that it seems odd to repeat the same code, but we can get away with that if these functions and classes are just utilities for internal use. What do you think?

tirthasheshpatel: Ok, I took a shot at it and removed almost all the duplicated code. Runs fine with a few warnings from `matplotlib` animation API.

tirthasheshpatel: Oh! The warning was because the solution initially is empty and matplotlib is not able to find the required arguments to draw... I think it's fine.

tirthasheshpatel:
@antmarakis is this ready to merge?

tirthasheshpatel:
oh, I see why the tests fail! totally forgot to add support for python 3.5. Lemme do that real quick!

omar-3:
I have been searching for something like that for a while and I cloned your repository to try them out, the A* algorithm works perfectly, but bfs/dfs doesn't find a solution and they got stuck despite them being complete algorithms. Am I missing something?

PR #1158, opened 2020-01-20
Added xgboost algorithm

AstikaNehra:


dmeoli:
Thanks @AstikaNehra. This branch is the same scattered grab-bag as #1171/#1172/#1177 (xgboost.ipynb, a pima-indians CSV, a 'chapter 20' directory with a space, duplicate adam-optimizer notebooks) and isn't a focused, mergeable change. xgboost is also a heavy external dependency that's out of scope for the core library. Closing; a single focused notebook from a clean branch would be welcome. Thanks!

PR #1163, opened 2020-01-28
added LinearRegressionLearner, LogisticRegressionLearner with tests and fixed NeuralNetLearner and PerceptronLearner

dmeoli:


deep_learning4e.py
```diff
@@ -338,64 +339,136 @@ def get_batch(examples, batch_size=1):
         yield examples[i: i + batch_size]
 
 
-def NeuralNetLearner(dataset, hidden_layer_sizes, l_rate=0.01, epochs=1000, batch_size=1,
-                     optimizer=stochastic_gradient_descent, verbose=None):
+class NeuralNetLearner(Learner):
```
antmarakis: I think this is not the best way to implement NNs. Even though you improved and streamlined the process a bit, right now it is not very useful. In your `fit` function you both initialize the model *and* train it. It would be best if these two processes were separate. That way, we could initialize the architecture, fit it, and then fit it again if we need to. So, all the initialization stuff should go to `__init__` and all the training stuff in `fit`.

deep_learning4e.py
```diff
@@ -338,64 +339,136 @@ def get_batch(examples, batch_size=1):
...
             layer_input = layer_out
 
         return layer_out.index(max(layer_out))
 
-    return predict
-
 
-def PerceptronLearner(dataset, l_rate=0.01, epochs=1000, batch_size=1,
-                      optimizer=stochastic_gradient_descent, verbose=None):
+class PerceptronLearner(Learner):
```
antmarakis: Same as the NN. Don't add the initialization stuff in `fit`.

deep_learning4e.py
```diff
@@ -338,64 +339,136 @@ def get_batch(examples, batch_size=1):
...
+                                          l_rate=self.l_rate, batch_size=self.batch_size, verbose=self.verbose)
+        return self
+
+    def predict(self, example):
+        layer_out = self.learned_net[1].forward(example)
         return layer_out.index(max(layer_out))
 
-    return predict
+
+if __name__ == "__main__":
```
antmarakis: I don't think having a main adheres to the rest of the repo's structure, I would prefer to have all that in the tests.

learning4e.py
```diff
@@ -453,139 +464,216 @@ def information_content(values):
...
+    def predict(self, example):
         """Find the k closest items, and have them vote for the best."""
-        best = heapq.nsmallest(k, ((dataset.distance(e, example), e) for e in dataset.examples))
-        return mode(e[dataset.target] for (d, e) in best)
+        best = heapq.nsmallest(self.k, ((self.dataset.distance(e, example), e) for e in self.dataset.examples))
+        return mode(e[self.dataset.target] for (d, e) in best)
 
-    return predict
 
+class LossFunction:
```
antmarakis: Why do loss functions need to be a class? I think the code would be much simpler if we kept them as functions.

learning4e.py
```diff
@@ -453,139 +464,216 @@ def information_content(values):
...
-    Linear classifier with hard threshold.
-    """
-    idx_i = dataset.inputs
-    idx_t = dataset.target
-    examples = dataset.examples
-    num_examples = len(examples)
 
-    # X transpose
-    X_col = [dataset.values[i] for i in idx_i]  # vertical columns of X
+class MeanSquaredError(LossFunction):
```
antmarakis: Same here, I think we already defined loss functions in learning as actual functions, I don't see why we need to have them as classes.

deep_learning4e.py
dmeoli: Sorry, it was a main that I used to test functions pre PR and I forgot to remove.

deep_learning4e.py
dmeoli: Idem

deep_learning4e.py
dmeoli: Ok Antonis, I thought about when you said and I reported the implementation of NeuralNet learner and PerceptronLearner as they were as they represent only examples of how a neural network could be composed :)

learning4e.py
dmeoli: I'm sorry, this class is necessary to fix the X and y matrices before and then pass the methods to Scipy minimization method! Anyway, I think it is right to have classes for the loss functions as for the activation functions that expose two methods, one that returns the cost of the function and the other the derivative (or, in this case, the Jacobian).

learning4e.py
dmeoli: Idem. Anyway, in the next PR I will try to prepare the neural network for the use of these more explanatory classes in order to make it more modular by separating the part of backpropagation from that of optimization. As it is now, even if it works, I think it is not very intuitive in a learning sense.

PR #1164, opened 2020-01-28
SGD with momentum

### LINKED ISSUE #1162: SGD with momentum

So-ham:
Resolves #1162 
This is how it looks!!
scr


antmarakis:
Hi, thanks for the contribution!

Even though it is really nice, there are some things I would like to bring up:

a) In the code, we do not seem to have SGD with momentum. Usually here we use the notebooks to show how the implementations work, describing the algorithm along the way. Also, remember that this is a repository for a book, so if an algorithm is not in the book, there is not much use in adding it here.

b) I am afraid the section you added is really short and doesn't properly describe the algorithm.

I suggest you first check a). That is, does momentum exist in either the codebase or the book? Then, you can move on to b).

So-ham:
Really appreciate the feedback!! Will keep the points in mind not only for
this pull request but also for other upcoming contributions I will be
making.
Thanks for your time.

dmeoli:
Thanks @So-ham. This adds a short conceptual markdown section on SGD with momentum (plus two images), with no accompanying code, a few typos, and a stale base. The repo's maintained optimizer material lives with the actual implementations in deep_learning4e.py (which already provides SGD and adam), so this standalone explanation adds limited value as-is. Closing for now - if you'd like to contribute a worked momentum example tied to the code, a focused PR would be very welcome. Thanks!

PR #1165, opened 2020-01-29
Updated search.ipynb

### LINKED ISSUE #1124: Possible issue in search.ipynb
Hello everyone. 

First, I would like to say thank you for AIMA, it is really a good library. However, I guess there is an issue in search.ipynb. In TSP problem example, the condition in **hill_climbing** should not it be:

```python
if problem.value(neighbor.state) >= problem.value(current.state):
```
instead of:
```python
if problem.value(neighbor.state) <= problem.value(current.state):
```
Because, the **TSP_problem.value()** method is negative (inverted path cost) and you use **argmax**, then I conclude that you should consider a best **neighbor** one that is greater then the **current**. Am I wrong? Thanks.

So-ham:
Resolves #1124 i.e when neighbour state value is greater than current state value then current state value should be over-written by neighbour state value

### DIFF
--- a/search.ipynb
+++ b/search.ipynb
@@ -2852,7 +2852,7 @@
     "            break\n",
     "        neighbor = argmax_random_tie(neighbors,\n",
     "                                     key=lambda node: problem.value(node.state))\n",
-    "        if problem.value(neighbor.state) <= problem.value(current.state):\n",
+    "        if problem.value(neighbor.state) >= problem.value(current.state):\n",
     "            current.state = neighbor.state\n",
     "        iterations -= 1\n",
     "        \n",


antmarakis:
Hey, I left a (belated) comment on the issue. What we need to do is leave the implementation as is and instead invert the path costs.

So-ham:
Correct me if I am wrong, but i feel that there isn't any issue really in
implementing hill_climbing algorithm. It's just that we have to make the
user aware (using comment/markdown) that the implementation of
hill_climbing is for inverted/negative path costs so that they can take apt
steps while defining a problem.
Hence,I think a simple md/comment for this matter will be good enough for
closing this issue!!

PR #1166, opened 2020-01-29
Updated search.ipynb

### LINKED ISSUE #1124: Possible issue in search.ipynb

So-ham:
resolves #1124  regarding confusion for negative path cost method used in hill_climbing algorithm


antmarakis:
Hi, thanks for the contribution! Please don't delete and reupload files, since it is hard to check. Instead, directly change the notebook code and then submit a PR.

Issue #1151, opened 2020-01-11
algorithms with no pseudocode


Okhaledzaki:
There are many algorithms in the book haven't a written pseudocode (stochastic hill climbing - local beam and its variants - ... ) . It would be beneficial, if they are to be implemented I guess ... should someone implementing them follow their understanding and surely add adequate tests and so on IF there are no "reliable" papers on these algorithms?

antmarakis:
I believe it would be great to have these algorithms, yeah. But the implementations should be bare-bones and quite simple, no complex algorithms. And there should be a note in the implementation that the code has no corresponding pseudocode. Maybe add the source.

So, sure, this would be an interesting project.

saraansh1999:
I would like to contribute to this project in the python module if it were to be initiated.

antmarakis:
Hi @saraansh1999, welcome!

If you want to get started with this, I suggest you can simply find an algorithm from the book with no pseudocode and implement it here :)

mehmehul:
I would like to work on this.

dmeoli:
This is an open-ended umbrella (implement book algorithms that lack `pseudocode()` entries — stochastic hill climbing, local beam search, variants, …). To make it actionable it should be split into one concrete good-first-issue per algorithm with a clear spec. Leaving open as a tracking issue; individual algorithm additions can reference it.

dmeoli:
Made this concrete for the local-search family it names. Added to `search.py` (AIMA Section 4.1), as simple, faithful implementations of the book's prose — each docstring notes there is no corresponding pseudocode and cites the source section:
- `stochastic_hill_climbing` — pick a random uphill neighbor
- `first_choice_hill_climbing` — generate random successors until one improves (good when there are many successors)
- `random_restart_hill_climbing` — best result over several random starts
- `local_beam_search` — keep the *k* best successors at each step

All bare-bones (the project is educational; efficiency is secondary, as @antmarakis noted), with tests. Closing the umbrella now that the algorithms it lists are implemented — any further no-pseudocode algorithm can be its own focused issue referencing this one.

PR #1170, opened 2020-02-02
added python implementation of beam search

mehular0ra:
Solving issue #1151 
Added a python implementation for beam search algorithm in python.
Will try adding more in the future.

### DIFF
--- a/beam_search.py
+++ b/beam_search.py
@@ -1,0 +1,41 @@
+import numpy as np
+
+from numpy import array
+from numpy import argmax
+
+ 
+# beam search
+def beam_search_decoder(data, k):
+	sequences = [[list(), 1.0]]
+	# walk over each step in sequence
+	for row in data:
+		all_candidates = list()
+		# expand each current candidate
+		for i in range(len(sequences)):
+			seq, score = sequences[i]
+			for j in range(len(row)):
+				candidate = [seq + [j], score * -np.log(row[j])]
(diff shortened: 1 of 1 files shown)


Okhaledzaki:
I'm not one of the maintainers, but you obviously haven't taken into account how the code is structured in the repository or have a sense about the style of the written algorithm. Also, don't use numpy, @antmarakis said that he wanted to be as minimalist as possible, because after all this repository is for educational purposes and the code is very cluttered. as for the **algorithm** itself:

- You should choose the best k states from both the current "frontier" and all the successors. This is important for detecting when the algorithm gets stuck (see next point).

- You did not implement a good stopping criterion to prevent the algorithm from going into an infinite loop when it gets stuck. It is a local search algorithm after all, it can get stuck in a local minimum. In other words, it should stop when the current nodes are already the best nodes compared to all the successors (no possible progress).

The algorithm could be done really elegantly with basic python methods for lists and heapq.
Sorry for the long comment and good luck.

dmeoli:
Thanks @mehular0ra. This implements a sequence-decoding beam search (over a matrix of per-step probabilities, with hardcoded demo data and tab indentation) as a standalone root script, rather than AIMA's local beam search over a Problem's state space. It doesn't fit the library as-is. If you'd like to contribute local beam search, adding it to search.py against the Problem interface would be the way. Closing for now — thanks!

PR #1185, opened 2020-04-21
fixed svm for not posdef kernel matrix, updated .travis.yml with Python 3.8 and added svr with r2 and accuracy metrics

dmeoli:


dmeoli:
@antmarakis don't forget about this PR! 🙂

PR #1194, opened 2020-07-12
TravelingSalesman problem class was added.

Haimzis:
I added a class of Traveling Salesman problem, 
and a heuristic based on sum of MST edges.


chinokenochkan:
what's an example input for initial and goal?

Haimzis:
Hi !
an example for initial and goal inputs can be: 
initial=(0), under the assumption that we start from node 0 in the graph. 
goal=(0, ... ,0)
while '...'  must be some sequence of all the possible numbers of nodes ids.
we can start from any node, and also from a partial path such as: (1, 0, 2)
and the goal for such an input must be: (1, 0, 2, .... , 1)
note that any number except the first and last in goal tuple must be unique.

PR #1196, opened 2020-09-10
Improved PriorityQueue

Ewindar:
Explaining 861853b:
If the first element of two tuples are equal, then python tries to compare the 2nd elements, which will throw a TypeError if the 2nd elements can't be compared.
This change prevents that Error from happening in that special case.


Ewindar:
reason behind 8c10ad1:
https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas

Issue #1198, opened 2020-10-09
AttributeError: 'int' object has no attribute 'path_cost'---Error in uniform_cost_search_graph from "search.py"


Tridu33:
run 
```python
all_node_colors = []
romania_problem = GraphProblem('Arad', 'Bucharest', romania_map)
display_visual(romania_graph_data, user_input=False, 
               algorithm=uniform_cost_search_graph, 
               problem=romania_problem)
```
then click visualize,it shows:

```python
AttributeError                            Traceback (most recent call last)
~\anaconda3\envs\aima3\lib\site-packages\ipywidgets\widgets\interaction.py in update(self, *args)
    254                     value = widget.get_interact_value()
    255                     self.kwargs[widget._kwarg] = value
--> 256                 self.result = self.f(**self.kwargs)
    257                 show_inline_matplotlib_plots()
    258                 if self.auto_display and self.result is not None:

D:\tridu33\Py\jupyternotebook\tmp\notebook.py in visualize_callback(visualize)
    969                 global all_node_colors
    970 
--> 971                 iterations, all_node_colors, node = algorithm(problem)
    972                 solution = node.solution()
    973                 all_node_colors.append(final_path_colors(all_node_colors[0], problem, solution))

<ipython-input-9-632128d9ca73> in uniform_cost_search_graph(problem)
      2     "[Figure 3.14]"
      3     #Uniform Cost Search uses Best First Search algorithm with f(n) = g(n)
----> 4     iterations, all_node_colors, node = best_first_graph_search_for_vis(problem, lambda node: node.path_cost)
      5     return(iterations, all_node_colors, node)

<ipython-input-8-497b2a10a430> in best_first_graph_search_for_vis(problem, f)
     56             elif child in frontier:
     57                 incumbent = frontier[child]
---> 58                 if f(child) < f(incumbent):
     59                     del frontier[incumbent]
     60                 if f(child) < incumbent: #https://github.com/aimacode/aima-python/pull/1084/commits/da39ea357317e806cdf8b2ca5a25f1e3514effe4

D:\tridu33\Py\jupyternotebook\tmp\utils.py in memoized_fn(obj, *args)
    415                 return getattr(obj, slot)
    416             else:
--> 417                 val = fn(obj, *args)
    418                 setattr(obj, slot, val)
    419                 return val

<ipython-input-9-632128d9ca73> in <lambda>(node)
      2     "[Figure 3.14]"
[...]

PR #1199, opened 2020-10-12
Fixed the issue: 'int' object has no attribute 'path_cost'

Gabriele91:
The issue was explained in https://github.com/aimacode/aima-python/issues/1198
I fixed it adding the "PriorityQueue.get_item" method, and calling it in that way:
```
best_first_graph_search_for_vis(....)
          ...
          incumbent = frontier.get_item(child)
          if f(child) < f(incumbent)
                  do stuff...
 ```
This fix works well for uniform_cost_search_graph and **also for A***.


dmeoli:
Thanks @Gabriele91. The issue this addressed (#1198) no longer reproduces on current master: best_first_graph_search_for_vis runs correctly (verified a uniform-cost search Arad->Bucharest returns Sibiu->Rimnicu->Pitesti->Bucharest, cost  418). The frontier handling was fixed in #1242 - incumbent = frontier[child] returns the stored priority value and f(child) < incumbent compares numbers, so the extra PriorityQueue.get_item isn't needed. Closing as superseded by #1242. Thanks for the contribution!

PR #1204, opened 2020-11-03
Update search.ipynb

DarrionM:


dmeoli:
Thanks @DarrionM. This changes frontier[child] to frontier.get_item(child) in best_first_graph_search_for_vis, but get_item isn't a PriorityQueue method and the underlying issue was already fixed in #1242 (frontier[child] returns the stored priority and the comparison works). Same as #1199. Closing as superseded. Thanks!

PR #1212, opened 2021-01-10
Corrected error in the way h2() is called

mbalduccini:
In class EightPuzzle of search4e.ipynb, method h() was calling h2() incorrectly: self was passed as argument rather than being the object on which h2() was called. Doing so caused an error with the current version of Jupyter that prevented the 8-puzzle example from running.


cbw5803:
Exactly. But It has been three years, does this repo admin still maintain the repo?

dmeoli:
Thanks @mbalduccini. The 8-puzzle h2 call was fixed in #1322: on current master EightPuzzle.h is 'def h(self, node): return self.h2(node)', so the example runs. This PR is also superseded content-wise and carries a large amount of re-executed notebook output (including committed ModuleNotFoundError tracebacks). Closing as superseded by #1322. Thanks for reporting it!

PR #1214, opened 2021-01-18
Update utils4e.py

pctablet505:
there was a typo in line 377 which made the formula for cross_entropy_loss Wrong.

line 320, 324 so that it looks similar to above function definitions.


dmeoli:
Thanks @pctablet505! Your fork branch was deleted, so we re-applied this fix (authorship preserved) and merged it in #1330. Closing in favor of #1330.

PR #1217, opened 2021-01-20
Update utils4e.py

pctablet505:
Better implementation of PriorityQueue to reduce time complexity of
 __contains__, __getitem__ from O(n) to (1) 
and for __delitem__ from O(n) to O(log n) amortized

### DIFF
--- a/utils4e.py
+++ b/utils4e.py
@@ -36,7 +36,7 @@ class PriorityQueue:
         elif order == 'max':
             self.f = lambda x: -f(x)
         else:
-            raise ValueError("""Order must be either 'min' or 'max'.""")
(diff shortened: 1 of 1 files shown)


dmeoli:
Thanks @pctablet505. The O(1) lookup idea is nice, but this implementation has correctness issues (e.g. self.length is never decremented on pop — the length -= 1 is dead code after the return — so __len__ is wrong, and the empty-pop raise is misplaced). utils4e.PriorityQueue is used by every 4e module, so a rewrite of it needs to be carefully correct and tested. The current implementation works; closing this version. If you'd like to revisit an O(1) PriorityQueue, a fresh, tested PR would be welcome. Thanks!

PR #1221, opened 2021-02-17
Updated games4e.py

pctablet505:
Changed some variable names to reduce confusions.


dmeoli:
Thanks @pctablet505. This is a very large change to games4e.py (+1293 / -708 across 4 files) described as variable renames, which makes it hard to review safely and likely conflicts heavily with the current code. A focused PR (one concern at a time, minimal diff) would be much easier to review and merge. Closing for now — thanks!

PR #1229, opened 2021-07-12
Changed rl.py to reinforcement_learning.py

JosueCom:
README has rl.py instead of reinforcement_learning.py


JosueCom:
It doesn't change much except for the name in used in README for reinforcement learning file

PR #1238, opened 2021-10-21
Add nqueens csp

larawehbe:


dmeoli:
Thanks @larawehbe. An N-Queens CSP already exists in csp.py as NQueensCSP, and the repo keeps problems in the relevant modules rather than standalone root scripts, so this csp-nqueens.py duplicates existing functionality. Closing as already covered — thanks!

PR #1239, opened 2021-10-24
Update agents.py with a Observer Base Class

scsanty:
Uses Matplotlib for graphical display of Environments.


dmeoli:
Thanks @scsanty. This PR includes a personal coursework file (W1_Lab_..._R00195638.py) and so looks like it was opened by accident. The Observer-base-class idea for agents.py would be welcome as a focused PR (agents.py only, from a clean branch). Closing for now — thanks!

PR #1250, opened 2022-02-01
Fixed keras imports by using tensorflow.keras instead of just keras

nik312123:


dmeoli:
Thanks @nik312123. deep_learning4e.py and perception4e.py both import cleanly on the current stack (Keras 3.14.1 / TF 2.21). With Keras 3 the canonical form is the standalone 'import keras' (which master already uses), rather than tensorflow.keras, so this change is no longer needed. Closing as obsolete - thanks!

PR #1258, opened 2022-06-20
Added Missionary-Cannibals Solver

arishimam:


dmeoli:
Thanks @arishimam. MissCannibals(Problem) is a nice classic problem, but the repo doesn't keep standalone problem scripts at the repo root - problems live in the relevant modules (e.g. search.py) or as notebook examples. If you'd like to contribute it, a focused PR integrating it that way would be very welcome. Closing for now — thanks!

PR #1261, opened 2022-09-04
fixed fixed_lag_smoothing

shengmincui:
I found that the fixed_lag_smoothing() function in the code probability.py does not implement the persistence of the arguments. We wrote the fixed lag smoothing algorithm as a class and implemented the persistence of f, B, and t.


dmeoli:
Thanks @shengmincui. This PR is a very large dump (+153k lines across 39 renamed chapter notebooks) rather than a focused fix to fixed_lag_smoothing, so it can't be reviewed or merged as-is. If fixed_lag_smoothing needs a fix, please open a small focused PR touching just probability.py. Thanks!

PR #1267, opened 2022-10-16
Update the attribute error associated with pad_sequences

EnthusiasticTeslim:
Found an error while running py.test. I figured out that the module 'keras.preprocessing.sequence' has no attribute 'pad_sequences' in keras 2.9. Thus, I updated the deep_learning4e.py to reflect the latest update in keras.

### DIFF
--- a/deep_learning4e.py
+++ b/deep_learning4e.py
@@ -6,7 +6,7 @@ import statistics
 import numpy as np
 from keras import Sequential, optimizers
 from keras.layers import Embedding, SimpleRNN, Dense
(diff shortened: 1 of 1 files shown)


dmeoli:
Thanks @EnthusiasticTeslim. On the current stack (Keras 3.14.1 / TF 2.21) deep_learning4e.py imports cleanly as-is: both 'from keras.preprocessing import sequence' and sequence.pad_sequences resolve and work. The version-specific import error this addressed no longer occurs, so closing as obsolete. Thanks for the contribution!

PR #1268, opened 2022-10-29
Implemented 8queens and n-color problems fitness function

mleb2ici:


dmeoli:
Thanks @leb2ic. This adds a standalone problem.py script with a main()/switch() demo; the genetic algorithm itself already lives in search.py (genetic_algorithm), and the repo doesn't keep standalone demo scripts at the root. A focused notebook or test demonstrating GA fitness functions for n-queens/graph-coloring would be a nicer fit. Closing for now — thanks!

PR #1276, opened 2023-05-28
modified iris.attrnames to iris.attr_names to resolve attribute error as required in the issue mentioned 

Parthiv31:


Parthiv31:
Please do check whether the requirement is met

PR #1280, opened 2023-08-03
value iteration show value

acmonstery:


dmeoli:
Thanks for this! Since it was opened, the repo has been restructured: the 3e/4e modules were converged into a single `aima` package and the notebooks moved under `notebooks/`, so this PR targets files that no longer exist (e.g. `deep_learning4e.py`, root `mdp.ipynb`) and still carries some WIP/debug bits (placeholder cells, a broken `keras.utils.data_utils` import). Closing as superseded by the restructure — but a fresh value-iteration visualization built on the current `aima.mdp` + `notebooks/` layout would be very welcome.

PR #1293, opened 2024-02-14
Update agents.py

Moreauju:


dmeoli:
Thanks @Moreauju. This targets the same XYEnvironment off-by-one that has since been fixed at the source in #1303: the default bounds are now (width-1, height-1), so is_inbounds is already correct as written. Applying an extra -1 here on top of that fix would double-correct and drop the last row/column (valid range would become [0, width-2]). #1303 also fixes random_location_inbounds, which this patch didn't cover. Closing as superseded by #1303 - thanks for spotting the issue!

PR #1298, opened 2024-06-16
Update: argument in optimizers.SGD from lr to learning_rate

MateoGE01:
Running the tests provide a failed test in deep_learning4e.py due to recent changes in Keras, the argument lr has been changed to learning_rate.


dmeoli:
Thanks @MateoGE01. This exact change (optimizers.SGD lr to learning_rate in deep_learning4e.py) is already present on master, so there is nothing left to apply here. Closing as already addressed - thanks for catching it.

PR #1299, opened 2024-08-04
Fix sgd = optimizers.SGD(lr=0.01) error

cbw5803:
Follow the new keras argument name 'learning_rate' instead of 'lr' which is deprecated now. Will raise error message if not change to the new name.


cbw5803:
This has just been requested by #1299, close to avoid repeat

PR #1300, opened 2024-08-04
Add Search Algorithm

UditNayak:
**Issue: #1155 - Add Animations for Search Algorithms**

**Context:**

This PR introduces animations for search algorithms, enhancing the user experience by providing a visual understanding of the algorithm's progression.

**Changes Introduced:**

1. **Animation Components:**
   - Developed an `AnimateProblem` class to handle animations for various search algorithms.
   - Implemented methods for generating random lines, calculating straight-line distances, and transposing matrices.

2. **Integration with Search Algorithms:**
   - Integrated the `AnimateProblem` class with different search algorithms such as A*, Weighted A*, BFS, DFS, UCS, and Best-First Search.
   - Utilized a dictionary to define the solvers and their corresponding f-value functions.

3. **User Interface and Visualization:**
   - Created a matplotlib figure for displaying the animations.
   - Added visual indicators for obstacles, explored nodes, the current node, the initial state, and the goal state.

4. **Performance Optimizations:**
   - Optimized the animation rendering to ensure smooth performance.
   - Minimized potential lag by efficiently handling the exploration and plotting of nodes.

### DIFF
--- a/notebook4e.py
+++ b/notebook4e.py
@@ -1,6 +1,8 @@
 import time
 from collections import defaultdict
 from inspect import getsource
+import heapq
+import random
 
 import ipywidgets as widgets
 import matplotlib.pyplot as plt
@@ -10,12 +12,14 @@ from IPython.display import HTML
 from IPython.display import display
 from PIL import Image
 from matplotlib import lines
+from matplotlib.animation import FuncAnimation
 from matplotlib.colors import ListedColormap
 
 from games import TicTacToe, alpha_beta_player, random_player, Fig52Extended
 from learning import DataSet
 from logic import parse_definite_clause, standardize_variables, unify_mm, subst
-from search import GraphProblem, romania_map
+from search import GraphProblem, romania_map, Problem, Node
+from utils import PriorityQueue
 
 
 # ______________________________________________________________________________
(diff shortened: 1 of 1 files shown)


dmeoli:
Thanks for this, @UditNayak — animated search on a grid is a genuinely nice addition for #1155. A few things to address before it can be merged: (1) It currently crashes on modern matplotlib: ax.scatter(*transpose(...)) raises 'missing x and y' whenever the point set is empty - this happens on the very first step because the root node's solution() is [] (and also if obstacles is empty). Please guard empty sets before scatter. (2) obstacles=random_lines(...) as a default argument is evaluated once at import time and shared across instances - please use obstacles=None and generate them inside __init__. (3) import heapq is unused (the frontier uses utils.PriorityQueue). (4) straight_line_distance duplicates utils.euclidean_distance - please reuse it. (5) Minor: use triple-quoted docstrings to match the repo style. Once it runs end-to-end on current matplotlib, happy to re-review!

dmeoli:
Thanks for the contribution! This adds the visualization to `notebook4e.py`, which has since been folded into `aima/notebook_utils.py` by the package restructure, and the cell as written breaks on current matplotlib (scatter over an empty transpose) with a mutable default evaluated at import. With the branch not maintainer-editable I can't rework it in place, so I'm closing it as superseded — a reworked version against `aima/notebook_utils.py` (and the `notebooks/` demos) would be a great follow-up if you'd like to reopen it that way.

PR #1310, opened 2025-04-22
Update README.md

YashTiwari-2330:


YashTiwari-2330:
add some feture

dmeoli:
Thanks for the PR, @YashTiwari-2330. This line has since been updated on master to a more accurate wording — it now reads: 'Move from Python 3.5 to 3.7, and on to modern Python (the code now runs on Python 3.9 and up)', matching the current CI (Python 3.9–3.12). Pinning a specific patch like 3.7.11 here wouldn't be accurate, so we're closing this as already addressed. Thanks for helping out!

